PR #10, opened 2024-01-25
fix: version template for zarf action

mjnagel:
`v` is getting stripped off incorrectly in https://github.com/defenseunicorns/uds-common-tasks/pull/7/files


mikevanhemert (approved):
👀

PR #28, opened 2024-02-08
chore: add renovate zarf preset

rjferguson21:


Racer159:
This likely won't be needed anymore after #35 (defer to @zachariahmiller on closing)

PR #29, opened 2024-02-09
feat: add registry-login task

MxNxPx:


tasks/setup.yaml
```diff
@@ -1,6 +1,35 @@
...
+            --password-stdin \
+            "${REGISTRY}" \
+            >/dev/null
+
+          if [ "${REGISTRY}" = "registry1.dso.mil" ]; then
+             uds zarf tools registry digest \
+               "${REGISTRY1_TEST_IMAGE}" \
+               >/dev/null || (sleep "${REGISTRY_RETRY_INTERVAL}"; exit 1)
+          fi 
+        maxRetries: 10
```
justin-o12, line 34: I honestly don't know:  do we need 10 retries here.  We could have almost a minute waiting on this to fail with a 5 second interval.  Could we fail sooner?  Maybe this is not a big deal in the long run but it could add up.

MxNxPx: @justin-o12 - this was a decision to implement a greater number of retries which can either be run more quickly or slowly by adjusting the retry interval.  at the moment there is no way to influence the number programmatically, so it is hardcoded until it can be parameterized.

justin-o12 (approved):

tasks/setup.yaml
mjnagel: Just to add my 2 cents - think this is OK since the scenario we're catching here is almost always short lived with registry1 dropping connections. I'd rather have my pipeline retry for 60 seconds and pass at the last second vs having to manually trigger a retry in GH 😁

README.md
```diff
@@ -25,23 +25,29 @@ There are multiple task files available in this repository with different object
 
 ### setup.yaml
 
-This task file includes a single task `k3d-test-cluster` which sets up a k3d cluster running the `uds-core-istio-dev` package. This provides a baseline cluster with Istio and the UDS Core Pepr capabilities necessary to test against.
+This task file includes two tasks:
+
+-`k3d-test-cluster` sets up a k3d cluster running the `uds-core-istio-dev` package. This provides a baseline cluster with Istio and the UDS Core Pepr capabilities necessary to test against.
+-`registry-login` performs a registry login and retry a max of 10 times at a desired retry interval (in seconds). It reads in environment variables from the system executing the task to use for the registry login parameters.
```
mjnagel, line 31: Okay with being succinct but should we call out the required envs here (ref publish workflow where we did this)?

tasks/setup.yaml
```diff
@@ -1,6 +1,35 @@
...
+  - name: REGISTRY1_TEST_IMAGE
+    default: "registry1.dso.mil/ironbank/opensource/defenseunicorns/zarf/zarf-agent"
+  - name: REGISTRY
+    default: "${REGISTRY}"
+  - name: REGISTRY_USERNAME
+    default: "${REGISTRY_USERNAME}"
+  - name: REGISTRY_PASSWORD
+    default: "${REGISTRY_PASSWORD}"
+  - name: REGISTRY_RETRY_INTERVAL
+    default: "${REGISTRY_RETRY_INTERVAL}"
```
mjnagel, line 11: Would love a real default here instead of the env so that we don't have to pass it in.

zachariahmiller: maybe im being dense but why are the defaults being set like this?

README.md
MxNxPx: added

tasks/setup.yaml
```diff
@@ -1,6 +1,39 @@
...
+        default: ${REGISTRY}
+        description: "Registry"
+      registryUsername:
+        default: ${REGISTRY_USERNAME}
+        description: "Registry username"
+      registryPassword:
+        default: ${REGISTRY_PASSWORD}
+        description: "Registry password"
+      registryRetryInterval:
+        default: ${REGISTRY_RETRY_INTERVAL}
```
mjnagel, line 24:
```suggestion
        description: "Registry"
      registryUsername:
        description: "Registry username"
      registryPassword:
        description: "Registry password"
      registryRetryInterval:
        default: "5"
```

So my leaning is to do this - drop the `default: ${...}` stuff and add a single default for the retry interval. We don't need these defaults if we have the assumption that this task will be always be imported (which I think is a safe assumption given that these are common tasks designed to be imported). Definitely call out any scenarios this wouldn't work with - I played with these tasks a bit but I know there was more stress testing by others.

README.md
```diff
@@ -25,23 +25,33 @@ There are multiple task files available in this repository with different object
 
 ### setup.yaml
 
-This task file includes a single task `k3d-test-cluster` which sets up a k3d cluster running the `uds-core-istio-dev` package. This provides a baseline cluster with Istio and the UDS Core Pepr capabilities necessary to test against.
+This task file includes two tasks:
+
+- `k3d-test-cluster` sets up a k3d cluster running the `uds-core-istio-dev` package. This provides a baseline cluster with Istio and the UDS Core Pepr capabilities necessary to test against.
+- `registry-login` performs a registry login and retry a max of 10 times at a desired retry interval (in seconds). It reads in environment variables from the system executing the task to use for the registry login parameters.  The environment variable inputs are:
```
mjnagel, line 31:
```suggestion
- `registry-login` performs a registry login and retry a max of 10 times at a desired retry interval (in seconds). It reads in variables for the task to use for the registry login parameters.  The variable inputs are:
```
Think we can drop environment? In testing I think `--set` works fine and would prefer to document assuming that is used?

mjnagel (commented):
Don't take a look at this until after the weekend - I might've made some wrong assumptions and tests along the way.

tasks/setup.yaml
mjnagel: Opened a separate thread on a similar issue still 👀. TLDR is I don't think we care about calling this directly since these are common-tasks and should always be included elsewhere.

mjnagel (approved):

PR #30, opened 2024-02-12
chore: switch to uds zarf

mjnagel:
Removes need for zarf on its own, switches to the incredible `uds zarf` instead.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -34,29 +34,12 @@
       ],
       "datasourceTemplate": "github-tags",
       "extractVersionTemplate": "^v(?<version>.*)$"
-    },
-    {
-      "depNameTemplate": "defenseunicorns/zarf",
-      "fileMatch": ["README\\.md"],
-      "matchStrings": [
-        "- Zarf: (?<currentValue>.*)"
-      ],
-      "datasourceTemplate": "github-tags",
-      "extractVersionTemplate": "^v(?<version>.*)$"
-    },
-    {
-      "depNameTemplate": "defenseunicorns/zarf",
-      "fileMatch": ["\\.*\\.ya?ml$"],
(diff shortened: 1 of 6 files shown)


renovate.json
```diff
@@ -34,29 +34,12 @@
...
-      "depNameTemplate": "defenseunicorns/zarf",
-      "fileMatch": ["README\\.md"],
-      "matchStrings": [
-        "- Zarf: (?<currentValue>.*)"
-      ],
-      "datasourceTemplate": "github-tags",
-      "extractVersionTemplate": "^v(?<version>.*)$"
-    },
-    {
-      "depNameTemplate": "defenseunicorns/zarf",
```
zachariahmiller, line 48: just a thought on this. Do we add the settings.json in this repo for the schema validation and then add the match sting we use on that for uds-core for zarf?

I realize we sort of end up "dependent" on zarf version still in that case, but its just for that one scenario. Alternatively, uds could include the zarf schema in its repo and mitigate the need.

zachariahmiller (commented):
See comment. Depending on if/how we address that otherwise lgtm

renovate.json
mjnagel: 🤔 currently we don't have the schema so not an issue in uds common tasks but its a good callout. Probably need to pull the zarf schema from uds-cli to handle if we want to drop dependency completely.

Maybe we could also add a task for `uds zarf prepare lint`, which should validate against the correct schema (noting that it doesn't actually handle the IDE feedback issue).

mjnagel: Going to resolve this thread, think it makes sense to follow up once the above uds-cli issue is resolved.

MxNxPx (approved):

PR #35, opened 2024-02-12
chore: add common renovate config

Racer159:
This is a PR to hold a common renovate config for use across UDS (this will improve management of this file across repositories).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,24 +1,24 @@
-# UDS Common Tasks
+# UDS Common
 
-This repository contains common tasks used in UDS Packages for setup, creation, deployment, and publishing of packages and bundles.
+This repository contains common configuration and tasks used in UDS Packages for management, setup, creation, deployment, and publishing of packages and bundles.
 
-## Usage
+## Task Usage
 
-These tasks are designed to be consumed via remote task includes and GitHub Actions (WIP). It is typically easiest to include these tasks in a single root level `tasks.yaml`. Includes should follow the standard remote include pattern documented by UDS CLI:
+The tasks are designed to be consumed via remote task includes and GitHub Actions (WIP). It is typically easiest to include tasks in a single root level `tasks.yaml`. Includes should follow the standard remote include pattern documented by UDS CLI:
 
 ```yaml
 includes:
   - deploy: https://raw.githubusercontent.com/defenseunicorns/uds-common-tasks/$TAG/tasks/deploy.yaml
 ```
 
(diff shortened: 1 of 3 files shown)


Racer159:
Here is a test renovate Dashboard: https://github.com/Racer159/uds-common/issues/3

Note it is now picking up podinfo from the `zarf.yaml` in the root.

Racer159:
And here is a test against git, helm, and oci charts: https://github.com/Racer159/uds-package-sonarqube/issues/4

zachariahmiller (commented):
I think this looks good. A couple of things: 

1. I don't think the base renovate config should extend all the other things. Specifically, in uds-core we group in a particular way so we wouldn't want group:all extended in the base config. For config:base, workarounds, replacements I'm actually not sure... probably fine. I ran into some weird behavior when testing uds-core and removed them.

2. Does this handle all the values files that might need to be parsed a specific way depending on how the values are organized? I see the helm-values is included at the top but IIRC that doesn't match everything in the values and we had to do custom regex matchers based on certain ways that images were defined in chart values

3. I think the match for // type comment is missing (settings.json) which we need to keep the schema versions in sync

Maybe a good smoke test of this whole config would be to fork uds-core, revert it to a ref from right before we started merging in renovate updates, extend it with this config and see if we are getting updates for everything (there should be for every package directly under src). Thoughts?

zachariahmiller:
FWIW if we're confident this has what we need for most single package repos then I'm also happy to push this through for now and then iterate on anything missing or weird cases.

Racer159:
Made those changes - made what IMO is a better regex than the `//` one for the settings stuff (it works without needing to think about it).

Here is a forked version of uds-core from 2 mos ago: https://github.com/Racer159/uds-core/issues/2 - is this what you would expect?

zachariahmiller:
Pure magic ♥️

I think my only concern with what's already in here at this point would be if the images will get evaluated and updated with the way this values file is written:
https://github.com/defenseunicorns/uds-core/blob/main/src%2Fneuvector%2Fvalues%2Fregistry1-values.yaml

I also think it would be acceptable to say that falls into a per-repo config since it's so specific/unique 😆

I also noticed, probably an oversight on my part originally, but as this config is we would still need to be able to get updated oci versions for zarf packages used in a `uds-bundle.yaml`. Even just based on the file match strings I think this is missing, but happy to be told I'm wrong.

zachariahmiller (approved):
This looks great. Specific chart edge cases can and probably should be handled in their respective repos. Thanks for working on this! I know I'll be happy to never deal with a from scratch renovate setup again. 😄 :shipit:

PR #40, opened 2024-02-13
chore: introduce a fix for release-please not firing on PR workflows

Racer159:
Testing to see if this is a viable workaround for release-please PRs.

### DIFF
--- a/.github/workflows/pull_request.yaml
+++ b/.github/workflows/pull_request.yaml
@@ -3,9 +3,8 @@ name: Test Tasks
 on:
   # This workflow is triggered on pull requests to the main branch.
   pull_request:
-  # A workaround for release-please not triggering PR workflows.
-  pull_request_review:
-    types: [submitted]
+    # milestoned is added here as a workaround for release-please not triggering PR workflows (PRs should be added to a milestone to trigger the workflow).
+    types: [milestoned, opened, reopened, synchronize]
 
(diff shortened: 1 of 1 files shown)


Racer159 (commented):
submitting a review!

zachariahmiller:
@Racer159  I dont think the condition matters. This doesnt work by design. The solution is using a github app. 

Happy to facilitate the test though, what do you need here?

mjnagel:
@zachariahmiller the hope was that the workflow would trigger because a real person is doing the milestone action I think vs the github_token adding it somehow. The current issue is that the workflow doesn't trigger on a push from the github_token, but suspicion/hope was that some other trigger (milestone in this case) would work, as long as the actor is not github_token. Might be a flawed assumption there, but most places this issue is listed specifically callout pushes from the github_token being the problem.

zachariahmiller:
Got it. Yeah... Happy to be wrong, but I think this would have been an identified workaround if it worked. Could we test this somewhere that is not actually used for things? 😅

I still think the correct solution to this and all similar problems is the GitHub app. It is unclear to me why we aren't just pressing to get that created.

mjnagel:
Confirmed this works on https://github.com/mjnagel/test/pull/48, with a release-please created PR.

mjnagel (approved):

PR #44, opened 2024-02-16
chore: add common workflows for reuse across repos

Racer159:
This PR adds common workflows to a folder in this repo for reuse across other repositories.


.github/actions/save-logs/action.yaml
```diff
@@ -1,5 +1,5 @@
 name: save-logs
-description: "Save debug logs"
+description: Save debug logs
 
 runs:
   using: composite
```
mjnagel, line 5: I know you didn't create this action, but we should add an input for a suffix on `name` for the upload-artifact step to match what is needed on other repos (if the goal is to make this action consumable in other repos).

mjnagel (commented):
LGTM overall.

.github/actions/save-logs/action.yaml
zachariahmiller: @mjnagel I updated this if you want to re-review/approve since i now have a commit on the branch

mjnagel (approved):

PR #46, opened 2024-02-20
chore: update @ renovate rule to be more brewey

Racer159:
This updates the common renovate rule for @ separated versions to be brew specific (since that is majority where it is used).


Racer159:
Tested here:

https://github.com/Racer159/uds-common/blob/main/config/renovate.json5
https://github.com/Racer159/uds-package-sonarqube/pull/3

zachariahmiller (approved):
Awesome thanks for putting this in! :shipit:

PR #47, opened 2024-02-20
fix: use workflow env for pr.title

zachariahmiller:
PR of quick change to your PR to avoid shell command injection. Following uds-core's example https://github.com/defenseunicorns/uds-core/blob/bf7c1d28e077cf52d4f765b50d7efb8ce5d60fff/.github/workflows/commitlint.yaml#L29


Racer159 (approved):
lgtm

PR #62, opened 2024-03-06
chore: create shared test action to be used from calling workflows

ericwyles:
## Description

Created github action 'test' to contain common steps to build and test a uds package. Updated 'setup' task to log in to ironbank. Updated test-deploy workflow to use ironbank login from the shared setup.

## Related Issue

Relates to https://github.com/defenseunicorns/uds-software-factory/issues/17

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Other (security config, docs update, etc)

## Checklist before merging

- [ ] Test, docs, adr added or updated as needed
- [ ] [Contributor Guide Steps](https://github.com/defenseunicorns/uds-package-gitlab/blob/main/CONTRIBUTING.md#developer-workflow) followed


.github/actions/setup/action.yaml
```diff
@@ -1,6 +1,14 @@
 name: Setup Environment
 description: UDS Environment Setup
 
+inputs:
+  username:
+    description: Registry1 user
+    required: true
+  password:
+    description: Registry1 PW
```
zachariahmiller, line 9: small nit: Can we just write out password? I get that the description is pretty much already redundant 😄

ericwyles: Done!

zachariahmiller (approved):

PR #63, opened 2024-03-07
feat: update uds-core to 14.5

corang:


mjnagel:
Looks like renovate was waiting on its schedule but it kicked off and is over on https://github.com/defenseunicorns/uds-common/pull/64/files

mjnagel (approved):

PR #68, opened 2024-03-13
chore: hotfix the renovate config to properly pickup git chart versions and registry 1 images

Racer159:


zachariahmiller (approved):
:shipit:

PR #70, opened 2024-03-15
chore: add a pull task to get the latest package release

Racer159:
This adds a common workflow to get the latest published package release for a given UDS Zarf package.


tasks/deploy.yaml
```diff
@@ -10,6 +14,22 @@ tasks:
...
+      - description: Get the current Zarf package name
+        cmd: cat zarf.yaml | yq .metadata.name
+        setVariables:
+          - name: PACKAGE_NAME
+      - description: Get latest tag version from OCI
+        cmd: uds zarf tools registry ls ghcr.io/defenseunicorns/packages/uds/${PACKAGE_NAME} | grep ${FLAVOR} | sort -V | tail -1
+        setVariables:
+          - name: LATEST_VERSION
+      - description: Deploy the latest package release
+        cmd: uds zarf package deploy oci://ghcr.io/defenseunicorns/packages/uds/${PACKAGE_NAME}:${LATEST_VERSION} --confirm --no-progress ${{ .inputs.options }}
```
zachariahmiller, line 31: No emojis for the oci ref? 😢

Racer159: Its a draft! Leave me alone!

Racer159: Actually it won't work :/

tasks/publish.yaml
```diff
@@ -1,6 +1,6 @@
 variables:
   - name: TARGET_REPO
-    default: oci://ghcr.io/defenseunicorns/packages/uds
+    default: ghcr.io/defenseunicorns/packages/uds
```
zachariahmiller, line 3: if the TARGET_REPO variables were actually getting passed down this would be a breaking change but as it stands today i believe no impact

Racer159: We could bump the version of this repo if you wanted - I did check and it isn't used since that is a very canonical location at this point.

zachariahmiller: More just noting. I think im fine with it since it will be a new tag anyways and technically it would be a future update to uds-cli that would cause the issue

zachariahmiller (approved):
:shipit:

tasks/pull.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+tasks:
+  - name: latest-package-release
+    inputs:
+      path:
+        description: Path relative to the repositories root where the package needs to go
+        default: .
+    actions:
+      - description: Get the current Zarf package name
+        cmd: cat zarf.yaml | yq .metadata.name
```
MxNxPx, line 15: wondering if there is a different way to do this that doesn't require `yq` perhaps?

.github/actions/test/action.yaml
```diff
@@ -24,3 +27,8 @@ runs:
     - name: Test package
       run: uds run test-package
```
ericwyles, line 28: question: Should this 'Test package' and the next 'Test Package' step be an either/or? Meaning this one should be "if not inputs.upgrade"?

ericwyles: I see how it works in the gitlab PR now. Resolving this!

PR #73, opened 2024-03-19
chore: add a publish job for dev bundles

Racer159:
Adds a publish task to publish dev versions of bundles for testing.

### DIFF
--- a/tasks/publish.yaml
+++ b/tasks/publish.yaml
@@ -28,8 +28,8 @@ tasks:
       - description: Publish bundles for the amd64 and arm64 architectures
         cmd: |
           set -e
-          uds publish bundles/uds-bundle-*-amd64-${VERSION}.tar.zst oci://${TARGET_REPO}/bundles --no-progress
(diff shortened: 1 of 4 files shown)


tasks/publish.yaml
```diff
@@ -1,16 +1,44 @@
...
+        description: Path to the UDS bundle being published
+        default: bundle
+      version:
+        description: The version of the bundle to publish
+        required: true
+    actions:
+      - description: Publish bundles for the amd64 and arm64 architectures
+        cmd: |
+          set -e
+          uds publish ${{ .inputs.path }}/uds-bundle-*-amd64-${{ .inputs.version }}.tar.zst oci://${TARGET_REPO}/bundles --no-progress
```
zachariahmiller, line 40: I think im totally fine with this being either way, but in line with making the version an input do you want to do the same with TARGET_REPO?

zachariahmiller: No is a totally valid answer 🤣

Racer159: So is YES :sweat_smile:

Racer159:
Here is what it would look like to implement this in another repo: https://github.com/defenseunicorns/uds-package-gitlab/pull/76/files

zachariahmiller (approved):

PR #75, opened 2024-03-20
chore: add an explicit architecture input

Racer159:
This adds an explicit architecture input to create tasks


tasks/create.yaml
```diff
@@ -25,5 +33,13 @@ tasks:
       config:
         description: File name of the bundle config file. Defaults to uds-config.yaml. Path is relative to the path input.
         default: uds-config.yaml
+      architecture:
+        description: The architecture of the bundle to create
+        default: ${UDS_ARCH}
     actions:
-      - cmd: UDS_CONFIG=${{ .inputs.path }}/${{ .inputs.config }} uds create ${{ .inputs.path }} --confirm --no-progress --architecture=${UDS_ARCH} ${{ .inputs.options }}
+      - cmd: |
+          if [ ${FLAVOR} != "registry1" ] || [ ${{ .inputs.architecture }} != "arm64"]; then
```
zachariahmiller, line 41: `"arm64" ]` instead of `"arm64"]`

Racer159: pls suggest

zachariahmiller (approved):
👍

PR #77, opened 2024-03-21
chore: add the ability to spoof to common

Racer159:
This adds spoofing into common to work around issues with pulled releases for upgrades while release please is trying to act on them.


zachariahmiller (approved):
🥈 fun stuff

PR #81, opened 2024-03-22
chore: hotfix the spoof containing a dash in the input and add a publish step

Racer159:
Fix this again...


zachariahmiller (changes requested):
Lets try adding "tests" to the tasks so the repo self-validates the changes. Will make it easier to iterate quickly with high confidence

zachariahmiller (approved):
LGTM!

PR #85, opened 2024-03-26
chore: remove schedule on renovate

mjnagel:
Open to discussion on this - personally dislike using a schedule on the shared config. I think that could be done per-repo if people desire. I think it creates more confusion around why a certain update is not in a PR already.

### DIFF
--- a/config/renovate.json5
+++ b/config/renovate.json5
@@ -10,7 +10,6 @@
     "reviewersFromCodeOwners": false,
     "timezone": "America/New_York",
     "rebaseStalePrs": true,
-    "schedule": ["after 7am and before 9am every weekday"],
     "dependencyDashboard": true,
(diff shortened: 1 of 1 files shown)


zachariahmiller:
@mjnagel I dont feel strongly either way. fwiw you can click the checkbox on the dashboard i think? I would be curious if adding a schedule at the per repo config level overrides the common config or doesn't work... If so then the issue feels moot and its just a "sensible" default. If not then happy to discuss changing. Thoughts @Racer159 ?

mjnagel:
@zachariahmiller yeah I think overriding at the repo level would work. I guess I'd prefer the default be no schedule and individual repos add the schedule if desired but can be overridden on that 😆. My gut says the default schedule is not intuitive, and while the checkbox does work - again not super intuitive? Okay to just close this out if the consensus is to not do this!

Racer159 (approved):

PR #86, opened 2024-03-26
chore: udpdate k3d-core-istio-dev to k3d-core-slim-dev

ericwyles:
Update setup task to deploy new k3d-core-slim-dev instead of k3d-core-istio-dev

Tested it locally with 'uds run setup:k3d-test-cluster' and verified I can 'zarf connect keycloak'

image


zachariahmiller (approved):
:shipit:

PR #88, opened 2024-03-26
chore: update registry login to happen in the common env setup action

Racer159:
This removes the `registry-login` task and instead moves it to the common env setup action.


zachariahmiller (approved):
🚢 🇮🇹

PR #91, opened 2024-03-27
chore: add upgrade tests to common

Racer159:
This adds upgrade testing to the workflows here to catch workflow issues earlier.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -41,12 +41,17 @@
     }
   ],
   "packageRules": [
-    { 
-      "matchFileNames": [".github/**"],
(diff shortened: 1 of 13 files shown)


renovate.json
```diff
@@ -41,12 +41,17 @@
     }
   ],
   "packageRules": [
-    {
```
zachariahmiller, line 44: thoughts on moving this to common instead of just removing? I think in most cases we would want renovate to pin digests for us if something was added without one.

zachariahmiller (approved):
:shipit:

zachariahmiller:
@Racer159 I cant fix the checks on this one

Issue #53, opened 2024-03-02
Add setup task for full uds-core and allow default, up to date, version of uds-core/core-istio to be overridden with a variable


zachariahmiller:
There should be a setup task for deploying full uds-core in addition to core-istio as for some jobs either will be needed.

By default we want the setup task to deploy the latest version, but should have the capability for someone to override that if they are working to get their dependent project up to date or need to test against an older version for some reason. 

```[tasklist]
### Tasks
- [x] Add task to deploy full uds-core
- [x] Add inputs to both the uds-core and core-istio tasks for the uds-core version
- [x] Move the version to a default variable, along with the renovate comment to trigger updates
- [x] Add note to the documentation about how to specify a certain version of uds-core or core-istio
```

mjnagel:
> By default we want the setup task to deploy the latest version, but should have the capability for someone to override that if they are working to get their dependent project up to date or need to test against an older version for some reason.

Not totally opposed to this, but I think there's some value in forcing people to use "latest" if they are consuming the latest uds-common tasks. It should encourage better behaviors around staying up to date, and there's still options if they want an older version (use an older uds-common or a local task). I think if we added it as an input we would need to be careful to ensure its tracked with renovate, otherwise this becomes one of those things that people just hit an issue with updates and leave it pinned. Ideally we could do that as part of the common renovate config - but I'm not sure how easy that would be to identify generically 🤔. The other concern I'd have... we don't expose any similar var for uds-cli version, and there's potential for those two to be coupled/dependent.

Just some food for thought - I can understand the desire for flexibility but trying to weigh that against encouraging good patterns/updating these dependencies, and simplifying that for the end user. But definitely want to hear other opinions - I'm not the main one consuming this workflow so curious on the end-user perspective.

zachariahmiller:
Problem is that's unrealistic if there are multiple changes in a couple days that could be breaking. These tasks are meant flexible and reusable for most use cases, not force people to have to reprioritize something that may be a priority for them. 

Also maybe something breaks in uds-core but a release here also had a needed change for a task that is blocking other work that needs to be done. This is only intended as an escape hatch to temporarily avoid maddening problems that could impact certain teams differently.

[...]

PR #95, opened 2024-03-29
chore: add a full uds task and inputs

Racer159:
This adds a full deployment for uds-core and options to setup versions.

Fixes: #53


zachariahmiller (approved):
:shipit:

PR #97, opened 2024-04-02
chore: add latest-bundle-release task

zachariahmiller:


tasks/pull.yaml
```diff
@@ -32,3 +32,31 @@ tasks:
...
+        setVariables:
+          - name: LATEST_VERSION
+      - description: Pull the latest bundle release
+        cmd: ./uds pull oci://${TARGET_REPO}/bundles/${BUNDLE_NAME}:${LATEST_VERSION} --no-progress -o ${{ .inputs.path }}
+      # TODO (@WSTARR): This is currently needed to get around the chicken+egg condition when release please updates the version in GH
+      - description: Spoof the latest release to the current version
+        cmd: |
+          if [ ${{ .inputs.spoof_release }} != "false" ]; then
+            CURRENT_VERSION=$(cat ${{ .inputs.path }}/uds-bundle.yaml | yq .metadata.version)
+            CURRENT_PKG_NAME=$(cat ${{ .inputs.path }}/uds-bundle.yaml | yq .metadata.name)
```
Racer159, line 60: do we want to use path both for where the bundle is going, and where the definition is?  Do we want to normalize the package part as well?

tasks/pull.yaml
```diff
@@ -32,3 +32,31 @@ tasks:
...
+      - description: Get the current uds bundle name
+        cmd: cat ${{ .inputs.path }}/uds-bundle.yaml | yq .metadata.name
+        setVariables:
+          - name: BUNDLE_NAME
+      - description: Get latest tag version from OCI
+        cmd: ./uds zarf tools registry ls ${TARGET_REPO}/bundles/${BUNDLE_NAME} | sort -V | tail -1
+        setVariables:
+          - name: LATEST_VERSION
+      - description: Pull the latest bundle release
+        cmd: ./uds pull oci://${TARGET_REPO}/bundles/${BUNDLE_NAME}:${LATEST_VERSION} --no-progress -o ${{ .inputs.path }}
```
Racer159, line 54: TARGET_REPO could probably be an input like publish is now

Racer159 (changes requested):

tasks/pull.yaml
```diff
@@ -32,3 +32,31 @@ tasks:
...
+      # TODO (@WSTARR): This is currently needed to get around the chicken+egg condition when release please updates the version in GH
```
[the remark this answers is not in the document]
zachariahmiller: lol you dont want the duplication of your todo to still reference you?

tasks/pull.yaml
zachariahmiller: For the first part, yes or at least that's how it is in every repo using bundles today. For the second part are you saying because the var is named PKG or normalize package to go to the same location as the zarf.yaml?

tasks/pull.yaml
zachariahmiller: i was keeping parity with the pull-package. fine with updating both

tasks/pull.yaml
zachariahmiller: i made the other changes. dont think we should change this.

zachariahmiller: fixed

Racer159 (approved):

PR #99, opened 2024-04-02
chore: fix renovate env var rule overmatching

Racer159:
tested here: https://github.com/Racer159/uds-common/pull/11


zachariahmiller (approved):
:shipit:

PR #100, opened 2024-04-02
chore: fix the renovate config github digest pinning

Racer159:
tested here: https://github.com/Racer159/uds-common/pull/11


Racer159:
You can also see that it has pinned the node action (I purposefully unpinned to test)
image

zachariahmiller (approved):
👍

PR #102, opened 2024-04-05
chore: add a UDS package CR to make the package better for testing

Racer159:
This adds a UDSPackage CR to the `podinfo` package in this repo to make it more suitable for testing things in `uds-core`

### DIFF
--- a/uds-package.yaml
+++ b/uds-package.yaml
@@ -1,0 +1,14 @@
+apiVersion: uds.dev/v1alpha1
+kind: Package
+metadata:
+  name: podinfo
+  namespace: podinfo
+spec:
+  network:
+    expose:
+      - service: podinfo
+        podLabels:
+          app: podinfo
+        host: podinfo
+        gateway: tenant
+        port: 9898
--- a/zarf.yaml
+++ b/zarf.yaml
@@ -6,17 +6,22 @@ metadata:
   # x-release-please-end
 
 components:
-  - name: podinfo-kustomize
+  - name: podinfo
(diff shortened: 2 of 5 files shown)


uds-package.yaml
```diff
@@ -0,0 +1,14 @@
...
+  namespace: podinfo
+spec:
+  network:
+    expose:
+      - service: podinfo
+        podLabels:
+          app: podinfo
+        host: podinfo
+        gateway: tenant
+        port: 9898
```
zachariahmiller, line 14: Since you are doing this can you add a wait for network on this port to test it works?

zarf.yaml
```diff
@@ -15,6 +15,8 @@ components:
         namespace: podinfo
         kustomizations:
           - github.com/stefanprodan/podinfo//kustomize?ref=6.6.1
+        files:
+          - uds-package.yaml
```
zachariahmiller, line 19: If this is also going to server as an example, do you want to put something not allowed in here as well and add an exemption?

zachariahmiller (changes requested):

uds-package.yaml
Racer159: done

zarf.yaml
Racer159: I don't think many people will use exemptions, but I did add an allow to show that off and since `echo` is a supported podinfo feature anyway.

zachariahmiller (approved):
:shipit:

PR #103, opened 2024-04-05
chore: pull the current bundle and package names when deploying

Racer159:
This pulls the specific package and bundle names and versions when running deploy tasks - this prevents errors when running uds tasks locally with older or different packages in place:
image


zachariahmiller (approved):
👍

PR #114, opened 2024-04-21
chore: update setup to uds 0.20.0

ericwyles:
Update setup task to install uds 0.20.0


zachariahmiller (approved):
Approving but consider just merging this one instead? 🤣 https://github.com/defenseunicorns/uds-common/pull/106

PR #129, opened 2024-05-15
chore: update package practices with a bit more feedback

Racer159:


docs/uds-package-practices.md
```diff
@@ -35,6 +35,13 @@ Below are the minimal services that a UDS package **must** integrate with, using
 - **May** end any generated secrets with `-sso` to easily locate them when querying the cluster.
 - **May** template Keycloak fields to provide flexibility for delivery customers to configure.
 
+### Prometheus
+
+- **Must** implement monitors for each application metrics endpoint using the `monitor` key as required.
```
mjnagel, line 40: I think you may have tried to capture this with `as required` but there's cases where you would want to use the upstream chart monitors. There may also be cases where you would make a monitor in the uds-config chart (we are intentionally not exposing the full complex servicemonitor spec at the moment, and would have to consider additions case-by-case for what fits the 80-90% without making things complex.

Not really sure what to suggest to clarify the wording but I think what I'm getting at is `**Must** implement monitors for each application metrics endpoint using <some method>`.

Racer159: changed the wording to just list those options. does that work?

mjnagel: Works for meeee

mjnagel (approved):

PR #135, opened 2024-05-24
chore: pull debug / log actions from uds-core

Racer159:
This pulls in the debug and log actions from `uds-core` that give more debugging power to CI runs.


ericwyles (approved):
I dig it!

PR #143, opened 2024-06-05
chore!: update publish to take architecture as an input

ericwyles:
Updates the `publish:package` and `publish:test-bundle` tasks to publish a single architecture. 

The architecture to publish defaults to [UDS_ARCH](https://github.com/defenseunicorns/uds-cli/blob/main/docs/runner.md#architecture-environment-variable) and can also be overridding by executing the task with the desired `architecture` specified as input.

This also updates the environment setup action to be able to install `uds-cli` on `arm64` linux runners.

⚠️ Breaking changes are included in this PR:

`publish:package` and `publish:test-bundle` will no longer default to publishing both `arm64` and `amd64` architectures, and it will no longer include conditional publish logic based on the flavor.

If you need to specify specific architecture(s) for the package you are publishing, you can do so as shown below. Note: If you need to publish multiple architectures, you will need to repeat the task for each one, specifying the architecture as an input for each.

``` yaml
      - description: Publish the package
        task: publish:package
        with:
          # x-release-please-start-version
          version: "9.8.0-uds.0"
          # x-release-please-end
          architecture: amd64

      - description: Publish the package
        task: publish:package
        with:
          # x-release-please-start-version
          version: "9.8.0-uds.0"
          # x-release-please-end
          architecture: arm64
```


ericwyles:
Tested the environment setup portion of this in these workflows here off of a branch of sonarqube: https://github.com/defenseunicorns/uds-package-sonarqube/actions/runs/9455905917/job/26047878979

.github/actions/setup/action.yaml
```diff
@@ -17,13 +21,22 @@ runs:
...
+      if: ${{ runner.arch != 'ARM64' }}
       shell: bash
-      # renovate: datasource=github-tags depName=defenseunicorns/uds-cli versioning=semver
-      run: brew install defenseunicorns/tap/uds@0.11.0
+      run: brew install defenseunicorns/tap/uds@${{ inputs.udsCliVersion }}
+
+    - name: Install UDS CLI
+      if: ${{ runner.arch == 'ARM64' }}
+      shell: bash
+      run: curl -o /usr/local/bin/uds -L https://github.com/defenseunicorns/uds-cli/releases/download/v${{ inputs.udsCliVersion }}/uds-cli_v${{ inputs.udsCliVersion }}_Linux_arm64 && chmod +x /usr/local/bin/uds
```
Racer159, line 39: would we be able to install the arm64 version by taking `runner.arch` and making it lowercase - GH doesn't have a native option for that (I don't think) but you could pass the arch into the `env` and use bash parameter expansion:

https://stackoverflow.com/questions/62627931/github-actions-expression-functions-string-manipulation

README.md
```diff
@@ -52,11 +52,27 @@ This task file includes two tasks:
...
+  - `TARGET_REPO`: the target OCI repository to publish the bundle (without the `oci://` scheme) - this defaults to `ghcr.io/defenseunicorns/packages/uds/bundles`.
+
+  The bundle will be matched based on the pattern `${PATH}/uds-bundle-*-${ARCHITECTURE}-${VERSION}.tar.zst` and published to `oci://${TARGET_REPO}`
+
 
-- `FLAVOR`: the flavor of the zarf package to publish, defaults to `upstream`.
-- `TARGET_REPO`: the target OCI repository to publish the zarf package to (without the `oci://` scheme) - this defaults to `ghcr.io/defenseunicorns/packages/uds`.
-- `VERSION`: the version of the zarf package to publish with no default. This should typically be version controlled by something like release-please.
 
 Note that this task will publish both `arm64` and `amd64` architectures, with the exception of the `registry1` flavor which will skip the `arm64` publish.
```
Racer159, line 77:
```suggestion
```
You could remove this note

Racer159 (changes requested):
looks good overall, just two notes

.github/actions/setup/action.yaml
ericwyles: I tried this but the amd64 runner.arch is actually "X64" and doesn't resolve. Just thinking we could do a ternary on it though and keep this to one command it skip installing brew.

ericwyles: Misread your comment but trying a couple things.

Racer159: Cool yeah I think that works and is much cleaner than the homebrew stuff.

Racer159 (approved):

PR #150, opened 2024-06-11
chore: upgrade uds cli-cli to v0.11.2

ericwyles:
Upgrade uds-cli to v0.11.2 because of an issue pulling image layers with v0.11.1

### DIFF
--- a/.github/actions/setup/action.yaml
+++ b/.github/actions/setup/action.yaml
@@ -11,7 +11,7 @@ inputs:
   udsCliVersion:
     description: The uds-cli version to install
     # renovate: datasource=github-tags depName=defenseunicorns/uds-cli versioning=semver
-    default: 0.10.4
+    default: 0.11.2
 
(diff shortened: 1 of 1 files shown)


.github/actions/setup/action.yaml
```diff
@@ -11,7 +11,7 @@ inputs:
   udsCliVersion:
     description: The uds-cli version to install
     # renovate: datasource=github-tags depName=defenseunicorns/uds-cli versioning=semver
-    default: 0.11.0
+    default: 0.10.4
```
Racer159, line 14: we can just upgrade to 0.11.2 now

Racer159 (changes requested):

.github/actions/setup/action.yaml
ericwyles: Done!

Racer159 (approved):

PR #157, opened 2024-06-21
feat(compliance): lula validate and evaluate capability definitions

brandtkeller:
## Compliance Tasks Addition

This update includes a `compliance.yaml` which contains two compliance tasks:

### Validate 

`validate` performs an assessment given a target `OSCAL` component definition. This produces an accompanying `assessment-results` OSCAL artifact capable of storing 1-> N results as historical artifacts. 

### Evaluate

`evaluate` consumes `assessment-results` OSCAL artifacts, compares the compliance of the latest result against an identified threshold, and returns/exits based upon state of compliance. Comparison here means validation that all findings (controls) previously identified as `satisfied` are still satisfied, and the identification of new passing controls in the event the latest result is more compliant that previously.

- If the compliance of the latest result is worse than the established threshold - exit non-successful with no change
- If the compliance of the latest result is equal to the established threshold - exit successful with no change
- if the compliance of the latest result is more compliant than the established threshold - exit successful - mark new result as threshold

### uds-common

The testing within this repository provides a reference and test that these tasks work as expected. The `test-upgrade` task executes two validations - setting the minimum results required for `evaluate` to execute. 

### Consumable UDS packages

Intent for deliverable uds packages would be a recommendation that the threshold be an artifact that is also stored within the repository. This bridges the gap in validation considering both changes to the package but also changes to the OSCAL artifacts (IE the input for assessments). 

When a new PR is submitted to a given package:
- an established (expected) threshold exists
- the validate task will produce an assessment-result for the proposed changes
- evaluate will determine compliance against established threshold
- (optional) pipeline orchestration should upload the combined assessment-results artifact to some location that is retrievable. In the event of greater-compliance, this artifact can be submitted back to the repository as the new threshold.

## Constraints

- Threshold identifier is waiting release (soon). This does not block execution - but improves the expected threshold identification
- Situation dictates _when_ to execute the validate. Given kubernetes applications it is recommended that validate occur after the pipeline has orchestrated standing up a testing environment and the deployment of proposed changes required for testing and validation purposes.


tasks.yaml
```diff
@@ -84,13 +85,17 @@ tasks:
...
       - task: test
+      - task: compliance:validate
 
   - name: test-upgrade
     description: Test an upgrade from the latest released package to the current branch
     actions:
       - task: create-podinfo-latest-release-bundle
       - task: setup:k3d-test-cluster
       - task: deploy:test-bundle
+      - task: compliance:validate
```
Racer159, line 96: we prolly don't need this here since it is just testing the previously released stuff again

.github/actions/save-logs/action.yaml
```diff
@@ -44,3 +44,4 @@ runs:
           /tmp/debug-*.log
           /tmp/uds-containerd-logs
           /tmp/k3d-uds-*.log
+          assessment-results.yaml
```
Racer159, line 47: I don't seem to see this file in the debug logs?
image
https://github.com/defenseunicorns/uds-common/actions/runs/9669308850?pr=157

Racer159 (changes requested):

tasks.yaml
brandtkeller: Great callout - we're including this for the following reason:

> The test-upgrade task executes two validations - setting the minimum results required for evaluate to execute.

This allows us to test the `evaluate` command which expects 2 or more results for comparison. The recommendation outside of this repository is to include the assessment-results as a threshold - but we did not see the need to include that here as an item that requires maintenance.

.github/actions/save-logs/action.yaml
brandtkeller: great point - changing the task use to include writing that file to tmp. Was previously mis-named.

tasks.yaml
Racer159: ah ok cool I missed that part.

Racer159 (approved):

PR #161, opened 2024-06-28
chore: add dockerfile `ADD` support

marshall007:
This allow `ADD` commands in Dockerfiles to be renovated. See screenshot for proof that it is working. Here is the related package PR: https://github.com/defenseunicorns/uds-package-mattermost/pull/99

image


Racer159 (approved):
saw the test through gather

PR #166, opened 2024-07-03
chore: fix commitlint workflow permissions

Racer159:
This fixes the permissions on the common `commitlint` workflow (to allow it to run on private / internal repos while we develop new things)


Racer159:
Tested here: https://github.com/defenseunicorns/uds-package-sigstore/actions/runs/9782603195/job/27009300573?pr=2

zachariahmiller (approved):

PR #173, opened 2024-07-12
fix: chainguard creds/renovate match

mjnagel:
Fixes the creds to update to latest token and remove hosttype since that causes issues with chainguard.


Racer159 (approved):
:monkey_face:

zachariahmiller (approved):
:shipit:

PR #175, opened 2024-07-12
chore: update CODEOWNERS with more specific permissions

Racer159:
This updates the `CODEOWNERS` file to protect some files but also open up more of the repository to others.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,12 @@
-/* @defenseunicorns/uds-core @defenseunicorns/swf
+/* @defenseunicorns/uds-core @defenseunicorns/swf @defenseunicorns/zarf @defenseunicorns/pepr @defenseunicorns/lula-dev @defenseunicorns/uds
+
+# Common Renovate Configuration (updates as soon as it hits `main`)
+/config/renovate.json5 @defenseunicorns/uds-core @defenseunicorns/swf @defenseunicorns/tech-leads
+
+# UDS Package Practices
+/docs/made-for-uds.svg @defenseunicorns/tech-leads
+/docs/uds-package-practices.md @defenseunicorns/tech-leads
+
+# Additional privileged files
+/CODEOWNERS @jeff-mccoy @austenbryan
+/LICENSE @jeff-mccoy @austenbryan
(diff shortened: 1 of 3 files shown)


CODEOWNERS
```diff
@@ -1 +1,12 @@
-/* @defenseunicorns/uds-core @defenseunicorns/swf
+/* @defenseunicorns/uds-core @defenseunicorns/swf @defenseunicorns/zarf @defenseunicorns/pepr @defenseunicorns/lula-dev @defenseunicorns/uds
```
mjnagel, line 1: Most of these groups are missing or don't have write access - should we rectify that before merging?

mjnagel: Might also be good to add the leapfrog team here? Not sure if uds-engine and/or uds-marketplace would be good as well?

Racer159: Yep will def do that - want to get agreement these are the right (write :sweat_smile:) groups first - good callout (forgot them)

Racer159: done

brandtkeller (approved):
Looks good pending repo setting modification for CODEOWNER permissions for file being valid.

corang (approved):

CODEOWNERS
```diff
@@ -1 +1,12 @@
-/* @defenseunicorns/uds-core @defenseunicorns/swf
+/* @defenseunicorns/uds-core @defenseunicorns/swf @defenseunicorns/zarf @defenseunicorns/pepr @defenseunicorns/lula-dev @defenseunicorns/uds @defenseunicorns/leapfrogai
+
+# Common Renovate Configuration (updates as soon as it hits `main`)
+/config/renovate.json5 @defenseunicorns/uds-core @defenseunicorns/swf @defenseunicorns/tech-leads
```
mjnagel, line 4: Probably makes sense to put `@defenseunicorns/leapfrogai` here too right?

mjnagel (approved):

CODEOWNERS
Racer159: I left it to the teams / people I know have regularly fought with renovate - we could add them but wanted to limit slightly for now.

PR #176, opened 2024-07-15
chore: fix version matching for UDS packages

Racer159:
This adds a version regex to the renovate for Zarf packages to match UDS package versioning schemes (as well as other semantic versions) - tested here:

- https://github.com/Racer159/uds-common/blob/d13629f5177feaf4257f8dcc5b6ea4a509077966/config/renovate.json5#L232C44-L234C129
- https://github.com/Racer159/uds-package-sonarqube/pull/17


mjnagel (approved):

zachariahmiller (approved):
:shipit:

PR #181, opened 2024-07-22
chore: fix the Zarf package renovate regex to the correct versionTemplate

Racer159:
This corrects `versioning` to `versioningTemplate` and adds compatibility to properly filter tag results.


Racer159:
Tested here: https://github.com/Racer159/uds-common/blob/main/config/renovate.json5#L233 / https://github.com/Racer159/uds-software-factory/pull/3

mjnagel (approved):

PR #183, opened 2024-07-23
chore: add additional install step to playwright install

Racer159:
This adds an additional action to ensure playwright deps are installed as per: https://github.com/microsoft/playwright/issues/7249#issuecomment-1597663397


oates (approved):
lgtm

mjnagel (approved):

PR #188, opened 2024-07-25
feat: add task for determining target repo based on flavor

marshall007:
Based on https://github.com/defenseunicorns/uds-core/pull/585


tasks/utils.yaml
```diff
@@ -0,0 +1,33 @@
...
+        default: "false"
+    actions:
+      - description: Determine repository for the given flavor/type of release
+        cmd: |
+          repo="${{.inputs.base_repo}}"
+          # unicorn flavor = private repository
+          if [ "${{.inputs.flavor}}" = "unicorn" ]; then
+            repo="${repo}/private"
+          fi
+          repo="${repo}/uds"
```
Racer159, line 25: this should be an input passed up and out of both pull / publish

Racer159 (changes requested):

tasks/utils.yaml
Racer159: for these people: https://github.com/defenseunicorns/leapfrogai/pkgs/container/packages%2Fleapfrogai%2Fleapfrogai-api

Racer159 (approved):

zachariahmiller (approved):
🚀

PR #191, opened 2024-07-31
feat: add action for assisting with badge verification and migrate to nginx

corang:
- [x] move badge logic to uds task
  - [x] conditional logic based on github ci or not
- [x] change example package to nginx
  - [x] all 3 flavors
  - [x] behind authservice or sso?
- [ ] extend badge logic to work for multi ns/package crd
- [x] Summary at end of badge script
- [x] More validations:
  - [x] package version matches app version in each flavor
- [x] group sections for gh actions (headers?)
- [x] better kubectl-validate pathing for binary


.github/actions/verify-badge/action.yaml
```diff
@@ -0,0 +1,257 @@
...
+      run: |
+        set -uo pipefail
+
+        # Validate UDS Package/Config chart is valid
+
+        echo "ℹ️ Installing kubectl-validate..."
+        go install sigs.k8s.io/kubectl-validate@latest
+
+        echo "ℹ️ Validating uds-config chart..."
+        kubectl validate <(helm template chart chart/)
```
Racer159, line 18: can we make chart an input (some packages are monorepos and while `chart` should be the default it isn't where it is everywhere.

.github/actions/verify-badge/action.yaml
```diff
@@ -0,0 +1,257 @@
...
+        echo
+
+        # Setup
+
+        NAMESPACE=$(yq '.components[].charts[].namespace' common/zarf.yaml | uniq)
```
Racer159, line 23: same with this - a decent example is monorepoed stuff like https://github.com/defenseunicorns/uds-package-sigstore or UDS Core itself.

.github/actions/verify-badge/action.yaml
```diff
@@ -0,0 +1,257 @@
...
+
+        echo
+
+        ## Should avoid workarounds such as disabling strict mTLS peer authentication.
+
+        PEERAUTH=$(kubectl get Peerauthentication -n "${NAMESPACE}" -o=json | jq -r '.items[].spec.mtls.mode')
+
+        if [ "$PEERAUTH" != "STRICT" ] && [ "$PEERAUTH" != "" ]; then
+          echo "❌ Peerauth is not strict"
+          exit 1
```
Racer159, line 72: we may not want to hard-exit on Shoulds - also this is tricky since some packages (like GitLab) place the PeerAuthentication exceptions in _other_ namespaces. https://github.com/defenseunicorns/uds-package-gitlab/blob/main/chart/templates/postgres-peerauthentication.yaml#L6

.github/actions/verify-badge/action.yaml
```diff
@@ -0,0 +1,257 @@
...
+
+        NETPOL_EXT_NOTKUBE_PORT_COUNT=$(echo "$UDS_PACKAGE_JSON" | jq -r '.spec.network.allow[] | select(.remoteGenerated != "IntraNamespace") | select(.remoteGenerated != "KubeAPI") | select(.ports != null or .port != null) | length' | wc -l)
+        echo "ℹ️ Non-IntraNamespace, non-KubeAPI network policies with ports: $NETPOL_EXT_NOTKUBE_PORT_COUNT"
+
+        if [ "$NETPOL_EXT_NOTKUBE_COUNT" -ne "$NETPOL_EXT_NOTKUBE_PORT_COUNT" ]; then
+          echo "❌ Not all applicable network policies are using ports"
+          exit 1
+        fi
+        echo "✅ All applicable network policies are using ports"
+
```
Racer159, line 115: We should probably alert on `remoteGenerated: Anywhere` as a warning

.github/actions/verify-badge/action.yaml
```diff
@@ -0,0 +1,257 @@
...
+        echo
+
+        # Network Policies
+
+        ## Must define network policies under the allow key as required.
+
+        NETPOL_AMOUNT=$(echo "$UDS_PACKAGE_JSON" | jq -r '.spec.network.allow | length')
+
+        if [ "$NETPOL_AMOUNT" -eq 0 ]; then
```
mjnagel, line 84: No netpols defined is actually acceptable...possibly. The UDS Operator will generate default deny policies and any necessary for istio, DNS, monitoring, etc. Some applications might not need more than that, so assuming the app is functioning this is an okay state.

.github/actions/verify-badge/action.yaml
```diff
@@ -0,0 +1,257 @@
...
+          exit 1
+        fi
+        echo "✅ Network policies are defined"
+
+        ## Should minimize network policies to specific selectors needed for Ingress/Egress traffic.
+
+        NETPOL_EXT_COUNT=$(echo "$UDS_PACKAGE_JSON" | jq -r '.spec.network.allow[] | select(.remoteGenerated != "IntraNamespace") | length' | wc -l)
+        echo "ℹ️ Non-IntraNamespace network policies: $NETPOL_EXT_COUNT"
+
+        NETPOL_EXT_SELECTOR_COUNT=$(echo "$UDS_PACKAGE_JSON" | jq -r '.spec.network.allow[] | select(.remoteGenerated != "IntraNamespace") | select(.selector != null) | length' | wc -l)
```
mjnagel, line 95: Might be missing where this is accounted for, but does this fail on remoteGenerated Anywhere/CloudMetadata/KubeAPI? I think those should be permitted here without failing.

.github/actions/verify-badge/action.yaml
mjnagel: There's a few other shoulds in the SSO section - agreed that they should not fail.

.github/actions/verify-badge/action.yaml
```diff
@@ -0,0 +1,257 @@
...
+          SSO_CLIENT_NAME=$(echo "$UDS_PACKAGE_JSON" | jq -r '.spec.sso[].name')
+          if [ "$SSO_CLIENT_NAME" != "${UDS_PACKAGE_NAME^} Login" ]; then
+            echo "❌ SSO client name not in correct format"
+            exit 1
+          fi
+          echo "✅ SSO client name is correct"
+
+          # Should clearly mark the client id with the group and app name uds-<group>-<application> (i.e. uds-swf-mattermost) to provide consistency in the Keycloak UI.
+          SSO_CLIENT_ID=$(echo "$UDS_PACKAGE_JSON" | jq -r '.spec.sso[].clientId')
+          if [ "$SSO_CLIENT_ID" != "uds-package-${UDS_PACKAGE_NAME}-${SSO_CLIENT_PROTOCOL}" ]; then
```
mjnagel, line 144: Package is not in the example format, may want to strike that and allow a regex or similar for that piece (i.e. could be SWF/LFAI/something else).

andrewg-xyz (commented):
Two overall suggestions
1) change everything to "warning" right now. There are assumptions made that may not be accurate.
2) shift logic to tasks.yaml - "builders" should be able to verify all of this easily prior to CI checkout.

Testing out with warnings on packages is a great start, but we should be wary of complication. Should some things shift to elsewhere in the stack? (like netpols from pepr for example or /healthz, ready endpoints)

.github/actions/verify-badge/action.yaml
zachariahmiller: Also postgres-operator, gitlab-runner and (soon) minio-operator config charts deploy things into multiple namespaces. For example there is a Package CR in the `gitlab-runner` and `gitlab-runner-sandbox` namespace. These are not referenced directly in the zarf component definition.

.github/actions/verify-badge/action.yaml
```diff
@@ -0,0 +1,257 @@
...
+        else
+          echo "❌ Tests folder does not exist"
+          exit 1
+        fi
+
+        echo
+
+        ## Should lint their configurations with appropriate tooling such as yamllint and zarf dev lint.
+
+        pre-commit run -a
```
zachariahmiller, line 257: we're not using pre-commit and had actually discussed removing the configs that were left over from copy/paste from older repos. If im not mistaken this would also require installing pre-commit, which isnt happening. IMO this should be removed. 

You could check for lint steps in the tasks and workflows instead of running pre-commit

.github/actions/verify-badge/action.yaml
```diff
@@ -0,0 +1,257 @@
...
+          echo "ℹ️ No monitors defined in the package, checking for ServiceMonitors"
+
+          # Check for built-in serviceMonitors
+          SERVICE_MONITOR_COUNT=$(kubectl get ServiceMonitor -n "${NAMESPACE}" -o json | jq -r '.items | length')
+          if [ "$SERVICE_MONITOR_COUNT" -eq 0 ]; then
+            echo "ℹ️ No ServiceMonitors defined, checking for PodMonitors"
+
+            # Check for built-in podMonitors
+            POD_MONITOR_COUNT=$(kubectl get PodMonitor -n "${NAMESPACE}" -o json | jq -r '.items | length')
+            if [ "$POD_MONITOR_COUNT" -eq 0 ]; then
```
zachariahmiller, line 182: So a couple things here. This actually is MUST (if possible)... For example we dont have monitoring on mattermost by default, despite a flag to enable it existing in the package because its locked behind a license. So its there, but would show up as not on this check.

Additionally, this doesnt actually check the intention of the rule, which is that ALL possible monitors are implemented. For example, gitlab has several. This would pass if only one was implemented, which is potentially misleading.

daveworth:
I wanted to publicly add that I love this work @corang - it's exactly the direction and shape of work I was hoping to see in Marketplace and it happened with exactly zero nudging from me.  ❤️ 🦄

.github/actions/verify-badge/action.yaml
anthonywendt: Annoying Nexus question. Its current client id is "uds:nexus" because it has to be a URI. Can we update this to be able to handle such a client ID.

mjnagel: @anthonywendt I think since this is a `should` it should be modified to just warn (not fail), which would cover Nexus and other apps that might require naming differently.

.github/actions/verify-badge/action.yaml
corang: Any suggestions on the best way to parse out all applicable namespaces?

.github/actions/verify-badge/action.yaml
corang: fixed the shoulds to be warnings

.github/actions/verify-badge/action.yaml
corang: Added!

.github/actions/verify-badge/action.yaml
corang: changed to a warning

.github/actions/verify-badge/action.yaml
corang: added an input and changed to a warning

corang:
> Two overall suggestions
> 
>     1. change everything to "warning" right now. There are assumptions made that may not be accurate.
> 
>     2. shift logic to tasks.yaml - "builders" should be able to verify all of this easily prior to CI checkout.
> 
> 
> Testing out with warnings on packages is a great start, but we should be wary of complication. Should some things shift to elsewhere in the stack? (like netpols from pepr for example or /healthz, ready endpoints)

1. I left some things as warning but the action will never "fail" due to continue on error
2. I'd like to do this eventually but while actively figuring out what logic goes in here I'm going to keep it as an action only

.github/actions/verify-badge/action.yaml
corang: Removed pre-commit, linting is already covered elsewhere in CI so I think it's fine

.github/actions/verify-badge/action.yaml
corang: Gave extra context in the messages on success

andrewg-xyz:
> > Two overall suggestions
> > ```
> > 1. change everything to "warning" right now. There are assumptions made that may not be accurate.
> > 
> > 2. shift logic to tasks.yaml - "builders" should be able to verify all of this easily prior to CI checkout.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Testing out with warnings on packages is a great start, but we should be wary of complication. Should some things shift to elsewhere in the stack? (like netpols from pepr for example or /healthz, ready endpoints)
> 
> 1. I left some things as warning but the action will never "fail" due to continue on error
> 2. I'd like to do this eventually but while actively figuring out what logic goes in here I'm going to keep it as an action only

@corang Isn't it easier to test as tasks you can run locally vs github actions? Copy on long-term desire. Keep in mind the in-portability aspect of doing this in github actions? What if a vendor is integrating and they dont' use github for example?

.github/actions/verify-badge/action.yaml
zachariahmiller: Not saying this is great or anything, but in my scenario, you could get the namespaces from the config chart helm templates. There are package CR's in any namespace that would need to be looked at in the examples we have today. So grabbing ns from Package resources out of the template. 

Then for @Racer159's case of mono repo like sigstore and uds-core, you could expand on that to check both `./chart` and `src/<package>/chart` where package is any folder under `src` 🤷

zachariahmiller:
@corang Was chatting with @Racer159 and think it might be better if the badge verification checks were written in go rather than bash. It would give far more flexibility and if produced as a standalone binary for now it could be easily integrated into our tooling (ie uds) in the future.

corang:
> @corang Was chatting with @Racer159 and think it might be better if the badge verification checks were written in go rather than bash. It would give far more flexibility and if produced as a standalone binary for now it could be easily integrated into our tooling (ie uds) in the future.

I agree, however I'm not the right person to do that, I'm way slower at writing things in go than bash. Can I move ahead with bash script for now and open an issue to migrate to go once this is merged?

chart/templates/uds-package.yaml
```diff
@@ -0,0 +1,56 @@
...
+      path: /metrics/
+      description: Metrics
+      podSelector:
+        app.kubernetes.io/name: nginx
+
+  sso:
+    - name: Nginx Authserviced
+      clientId: uds-package-nginx
+      redirectUris:
+        - https://nginx.uds.dev/login
```
andrewg-xyz, line 22: Should probably be a variable and ideally we could derive the base URL?

chart/values.yaml
```diff
@@ -0,0 +1,18 @@
+domain: "###ZARF_VAR_DOMAIN###"
+
+# customNetworkPolicies:
```
andrewg-xyz, line 3: Are these commented on purpose? suggest removing so there isn't confusion with what UDS package CR would provide.

docs/development-ide-configuration.md
```diff
@@ -1,22 +1,22 @@
 # IDE Configuration for developing UDS Packages
 
-This document explains how to configure specific IDEs to improve your development experience when developing UDS packages, such as setting up schema validation for Zarf packages, UDS bundles and Maru tasks. 
+This document explains how to configure specific IDEs to improve your development experience when developing UDS packages, such as setting up schema validation for Zarf packages, UDS bundles and Maru tasks.
 
 ## VSCode
 
-To configure your VSCode for schema validation, you must install a yaml language support [extension](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml). From there you have three options. You can configure VSCode globally, on a per repo basis or per file basis to provide validations. 
+To configure your VSCode for schema validation, you must install a yaml language support [extension](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml). From there you have three options. You can configure VSCode globally, on a per repo basis or per file basis to provide validations.
```
andrewg-xyz, line 7: What is our preference? imo per file basis is the most foolproof

docs/development-ide-configuration.md
```diff
@@ -1,22 +1,22 @@
...
 
 ### Global / Workspace Config
```
andrewg-xyz, line 9: Consider redirecting to external documentation, we probably don't want to get in the game of documenting other products.

docs/development-ide-configuration.md
corang: out of scope for this PR, got caught in a delete trailing whitespace command I ran

docs/development-ide-configuration.md
corang: out of scope for this PR, got caught in a delete trailing whitespace command I ran

chart/values.yaml
corang: These are commented on purpose, we're doing this across all packages as an example of how to use the custom net pols value

chart/templates/uds-package.yaml
corang: Missed that, I'll change it to reference values.domain

docs/development-ide-configuration.md
zachariahmiller: def out of scope of @corang's PR. FWIW we agreed to do this when we removed the .vscode/settings.json and `# yaml-language-server: $schema=https://raw.githubusercontent.com/defenseunicorns/zarf/main/zarf.schema.json` from all of the packages and the zarf.yamls because we should not be assuming that people are using vscode.

andrewg-xyz (approved):

Racer159 (changes requested):

.github/actions/verify-badge/action.yaml
```diff
@@ -0,0 +1,34 @@
...
+  using: composite
+  steps:
+    - name: Setup Go environment
+      uses: actions/setup-go@v5.0.2
+      with:
+        go-version: 1.22
+        check-latest: true
+
+    - name: Helm tool installer
+      uses: Azure/setup-helm@v4.2.0
```
zachariahmiller, line 25: these should be pinned, but renovate should do it for you iirc once there is an update to the versions 🤔

chart/.helmignore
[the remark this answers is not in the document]
corang: No, I'll delete

tasks/badge.yaml
```diff
@@ -0,0 +1,375 @@
...
+
+          if [[ "${COMMON_ZARF}" == "true" ]]; then
+            NAMESPACE=$(yq '.components[].charts[].namespace' common/zarf.yaml | uniq)
+          else
+            gh_warning "  ⚠️  There is no common zarf.yaml file"
+            NAMESPACE=$(yq '.components[].charts[].namespace' zarf.yaml | uniq)
+          fi
+          gh_notice "  ℹ️  Namespace: $NAMESPACE"
+
+          UDS_PACKAGE_JSON=$(kubectl get Packages -n "${NAMESPACE}" -o jsonpath='{.items[]}')
```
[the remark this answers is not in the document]
corang: I know and while this is written in bash I don't think its reasonable to attempt to handle multiple, I'll make sure its a goal in the refactor to go

.github/actions/verify-badge/action.yaml
corang: Yeah I assume renovate will take care of this

common/zarf.yaml
```diff
@@ -0,0 +1,31 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/zarf-dev/zarf/main/zarf.schema.json
+kind: ZarfPackageConfig
+metadata:
+  name: nginx
+  description: UDS nginx Common
```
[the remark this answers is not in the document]
corang: fixed

common/zarf.yaml
```diff
@@ -0,0 +1,31 @@
...
+      - name: nginx
+        version: 0.1.0
+        namespace: nginx
+        localPath: ../src/nginx
+        valuesFiles:
+          - ../values/common.yaml
+    actions:
+      onDeploy:
+        after:
+          - description: Validate nginx Package
```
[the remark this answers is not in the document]
corang: done

chart/.helmignore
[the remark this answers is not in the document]
corang: Nope, was a carryover from chart creation

src/nginx/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "1.16.0"
```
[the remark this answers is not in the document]
corang: done

src/nginx/templates/NOTES.txt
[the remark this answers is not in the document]
corang: Nope, was a carryover from chart creation

tasks.yaml
```diff
@@ -58,31 +72,26 @@ tasks:
...
+    description: Test and validate cluster is deployed with nginx
     actions:
       - wait:
           cluster:
             kind: Deployment
-            name: podinfo
-            namespace: uds-podinfo
+            name: nginx
+            namespace: nginx
             condition: available
```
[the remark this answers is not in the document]
corang: done

tasks/badge.yaml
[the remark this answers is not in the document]
Racer159: ok with this for now but created an issue to ensure it doesn't get lost: https://github.com/defenseunicorns/uds-common/issues/219

tasks/badge.yaml
```diff
@@ -0,0 +1,375 @@
...
+          fi
+
+          gh_endgroup
+
+          gh_group "🔍 Versioning Validation"
+
+          # Must be consistently versioned across flavors - this can take many forms but flavors should differ in image bases/builds not application versions.
+
+          ZARF_PKG_VERSION=$(yq '.metadata.version' zarf.yaml | cut -d '-' -f1)
+          IMAGE_WITH_ZARF_VERSION_COUNT=$(cat zarf.yaml | yq '.components[].images' | grep -o "${ZARF_PKG_VERSION}" | wc -l)
```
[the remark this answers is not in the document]
Racer159: Another reason not to have this - this will fail on renovate tasks where temporarily the version of the package is different from that of the images - IMO package versioning is more of a release engineering process thing than an artifact of the package itself (i.e. you could lookup releases and validate that things matched but even then it is not always true: https://github.com/defenseunicorns/uds-package-sigstore/blob/main/zarf.yaml#L7

.github/actions/verify-badge/action.yaml
Racer159: would prefer to default to do this if we notice it since technically if that repo was compromised our CI could be as well - in this repo it wouldn't give much more than the CG read creds but would still prefer to update prior to merge / adoption.

tasks/badge.yaml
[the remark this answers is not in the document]
corang: While it will "fail" at least for now this is a tool for people to use to help with package dev ans verification, the ci never actually fails it just prints info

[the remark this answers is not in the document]
corang: I still think there's value if someone tries to version a package with the chart again for instance or makes a typo in the initial version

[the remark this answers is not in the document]
Racer159: If we wanted to make this fail in the future though this check wouldn't work still though - in the very least it should be only a warning IMO - some apps will need to ignore app versioning and will always "fail" and every app would see that failure on releases.  Since this is a callable workflow once it is enforced all of the downstream repos will start having issues once they pick it up.

values/common.yaml
```diff
@@ -0,0 +1,30 @@
+securityContext:
```
zachariahmiller, line 1: Should we codify a naming pattern for values files now that its being changed/deviated from here? In all of SWF and uds-core the values files are named similarly and I guess that has changed recently in some of the other repos (like artifactory). I understand *assuming* someone has the visual context of the `values` directory the `-values` suffix is kind of redundant (less so if their view doesnt have that context), but would greatly prefer that we align the way we name things across repos/packages.
image
image
image
image
image

tasks/badge.yaml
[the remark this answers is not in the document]
zachariahmiller: Enough apps include additional third party images versioned in their own scheme, even if if they version their app images the same as the chart.yaml appVersion. Given that i dont think this check really makes sense.

values/common.yaml
corang: added `-values`

tasks/badge.yaml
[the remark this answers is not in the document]
corang: changed to a warning for now

corang:
Added a new issue https://github.com/defenseunicorns/uds-common/issues/220

Racer159 (approved):
lgtm with those two tech debt issues

zachariahmiller (approved):

PR #193, opened 2024-08-02
feat: add support for uds-core snapshots

Racer159:
This PR allows for specifying `uds-core` snapshots in a given repo in case there is an upcoming `uds-core` feature you are relying on.


zachariahmiller (approved):

tasks/setup.yaml
```diff
@@ -5,25 +5,31 @@ tasks:
...
+      snapshots:
+        description: Whether to pull from the nightly snapshot repository
+        default: "false"
       insecure_keycloak_admin:
         description: Automatically set a keycloak admin username / password
         default: "true"
     actions:
       - description: Create k3d cluster with slim UDS Core
-        cmd: ./uds deploy oci://defenseunicorns/uds/bundles/k3d-core-slim-dev:${{ .inputs.version }} --set INSECURE_ADMIN_PASSWORD_GENERATION=${{ .inputs.insecure_keycloak_admin }} --confirm --no-progress
+        cmd: ./uds deploy oci://defenseunicorns/uds/${{ if eq .inputs.snapshots "true" }}snapshots/${{ end }}bundles/k3d-core-slim-dev:${{ .inputs.version }} --set INSECURE_ADMIN_PASSWORD_GENERATION=${{ .inputs.insecure_keycloak_admin }} --confirm --no-progress
```
mjnagel, line 16: I think this is okay, although we maybe could also filter on `version` = some regex that matches date string + commit sha to auto-detect if its a snapshot or not (or just check valid semver vs not)? Then wouldn't need someone to pass in snapshot explicitly, just a version?

Racer159: IMO that feels clunky - I think this is simple enough (and unlikely to be used a ton - though I may eat my words there) - gonna merge as is and we can change it later

PR #196, opened 2024-08-05
chore: fix codeowners

mjnagel:
See https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file

/* only protects the top level files.


austenbryan (approved):
Remove me from all responsibilities of a real human and add a responsible guy named Dave

PR #197, opened 2024-08-05
feat: adds some renovate to find the jmx repo and tag for atlassian

Michael-Kruggel:
To be used by both atlassian packages to find the jmx exporter image necessary for metrics to work


mjnagel:
Just noting since this was discussed outside GitHub - this change did not feel common enough to bring in here. Might be worth future discussion on what meets the "common" bar but for now this one felt too specific to just 2 particular package(s).

PR #200, opened 2024-08-05
fix: update renovate creds

mjnagel:
For https://docs.renovatebot.com/getting-started/migrating-secrets/


Racer159 (approved):
Approving but super nit: Git_H_ub

PR #206, opened 2024-08-06
feat!: changed latest-package behavior

Michael-Kruggel:
This changes the behavior of the test upgrade task to use new logic. The reason for this is multifaceted:

1. The current task fails whenever running `test-upgrade` if the flavor does not exist beforehand. [EXAMPLE](https://github.com/defenseunicorns/uds-package-jenkins/actions/runs/10101198890/job/27934234195?pr=47)
2. If the bundle definition is changed (i.e. adding or changing the name of a helm override), the upgrade test will fail. [EXAMPLE](https://github.com/defenseunicorns/uds-package-jenkins/actions/runs/10098195291/job/27924700368?pr=45)
3. Because this test fails so often, it doesn't provide any meaningful value because we just override it and merge anyways.

This solution solves the above problems by cloning the most recent tag and building/deploying the package as it was, then using the PR branch to build its package and deploy over it. Also solves the non-existing flavor upgrade issue because if the flavor specified doesn't exist in the previous tag, it writes a warning to the log and moves on.


corang (approved):

zachariahmiller:
While i agree the upgrade could be better (we shouldnt have to spoof release), I dont think we should merge this in. This adds time to the upgrade jobs, which are already the longest running jobs and is just to mitigate a one time issue. If its not a one time issue, then there is something wrong with the job matrix in your workflow. 

If you have a flavor that does not have an upgrade available to run from OCI, just omit flavor-upgrade from the job in your specific project. Here is an example:


```yaml
  run-test:
    name: ${{ matrix.type }} ${{ matrix.flavor }}
    runs-on: uds-swf-ubuntu-big-boy-8-core
    timeout-minutes: 25
    strategy:
      matrix:
        flavor: [upstream, unicorn, registry1]
        type: [install, upgrade]
        exclude:
          - flavor: unicorn
            type: upgrade
```

corang:
@zachariahmiller 

> This adds time to the upgrade jobs, which are already the longest running jobs

Compared to the last merged PR in this repo it added 39 seconds, to the total job time, but the actualy upgrade piece was 31 seconds shorter, the clean runner action is being slow for some reason on this PR. It may take a bit longer due to having to create the archive for larger packages but i would assume its minimal from the comparisons above.

> is just to mitigate a one time issue. If its not a one time issue, then there is something wrong with the job matrix in your workflow.

It's not though, some facets of why we would skip it are, but the failing if the bundle changes can happen any time if a package maintainer changes how the bundle functions (which I don't think is that uncommon)

>just omit flavor-upgrade from the job in your specific project

Norming excluding jobs via commit to the repo feels like a bad practice that will end up with permanently skipped ci

Michael-Kruggel:
@zachariahmiller 
> This adds time to the upgrade jobs, which are already the longest running jobs.

Upgrade tests are supposed to take a long time and there is no way around that. When comparing previous upgrade tests, the only action that took longer was clean-runner for some reason. [Example - Previous upgrade test](https://github.com/defenseunicorns/uds-common/actions/runs/10288401000/job/28473729808?pr=207) [Example - This PR](https://github.com/defenseunicorns/uds-common/actions/runs/10293544356/job/28490039067?pr=206)

> and is just to mitigate a one time issue. If its not a one time issue, then there is something wrong with the job matrix in your workflow.

This also fixes if any changes are made to the local bundle or zarf configurations. This would also solve the problem I am having with upgrading unicorn flavors.

> If you have a flavor that does not have an upgrade available to run from OCI, just omit flavor-upgrade from the job in your specific project.

I disagree with the process of adding a bandage to a PR for it to then be removed down the line because the test doesn't function properly.

zachariahmiller:
Bundle failures across version updates are real problems that would need to be worked through by bundle creators who consume these changes. We try to make them work across changes at the package level to mitigate this. 

I'm not worried about the upgrade time for this repo (pod info 😅) but rather something like gitlab.

corang:
> Bundle failures across version updates are real problems that would need to be worked through by bundle creators who consume these changes
We agree here, the problem is that the current upgrade test doesn't properly do this

zachariahmiller:
> > Bundle failures across version updates are real problems that would need to be worked through by bundle creators who consume these changes
> 
> We agree here, the problem is that the current upgrade test doesn't properly do this


The upgrade test exists in the package repo, not uds-common. Common just executes it.

Racer159:
Thinking through the UX of noticing an issue with upgrades:

Known Good (what an upgrade "should" look like):
image

Failure Mode (as written in this PR (i.e. a tag doesn't get picked properly)):
image

Failure Mode (as @zachariahmiller suggested (i.e. someone forgets to turn the matrix back on):
image

IMO the second is more noticeable to a human after the fact.  If we wanted to make it more obvious/implement skipping I would prefer to do the skipping GitHub side:
image

Also to the slowdown issue - it would be possible and faster to do this with lookups on OCI tags instead of cloning the repo down.

zachariahmiller:
> I disagree with the process of adding a bandage to a PR for it to then be removed down the line because the test doesn't function properly.


I would argue it's not a bandage. It's correct because it doesn't exist. The same way it wouldn't make sense for me to have `registry1` in my matrix as a flavor if it didn't exist as a flavor

corang:
@Racer159 

> IMO the second is more noticeable to a human after the fact. If we wanted to make it more obvious/implement skipping I would prefer to do the skipping GitHub side

I like this better, I like GitLab where it says passed with warnings and I forgot GitHub doesn't do that, skipping would be much better

> Also to the slowdown issue - it would be possible and faster to do this with lookups on OCI tags instead of cloning the repo down.

This necessitates always publishing the test bundle for every package, should we be doing that?

Michael-Kruggel:
> > I disagree with the process of adding a bandage to a PR for it to then be removed down the line because the test doesn't function properly.
> 
> I would argue it's not a bandage. It's correct because it doesn't exist. The same way it wouldn't make sense for me to have `registry1` in my matrix as a flavor if it didn't exist as a flavor

I'm not sure how it is anything other than a bandage. If I add the unicorn flavor to any repo going forward, I need to this along with it. I then need to remember to remove this bandage after a release is cut for said flavor. It's temporary fix for a problem not currently addressed.

Racer159:
> @Racer159
> 
> > IMO the second is more noticeable to a human after the fact. If we wanted to make it more obvious/implement skipping I would prefer to do the skipping GitHub side
> 
> I like this better, I like GitLab where it says passed with warnings and I forgot GitHub doesn't do that, skipping would be much better
> 
> > Also to the slowdown issue - it would be possible and faster to do this with lookups on OCI tags instead of cloning the repo down.
> 
> This necessitates always publishing the test bundle for every package, should we be doing that?

It shouldn't req that - removing the spoof would require that but it could follow the skip logic based on package version - IMO if you are changing configuration to the point that it is hard breaking that should be somewhat painful because what you are doing will be painful to delivery and that pain forces the package creator to think about ways to handle that upgrade properly across bundle versions.

Michael-Kruggel:
> I'm not worried about the upgrade time for this repo (pod info :sweat_smile:) but rather something like gitlab.

So I thought about this and tested gitlab after seeing your first message. Locally these were the times I got:
  - Gitlab with my PR: 16:35
  - Gitlab with as is test: 16.22

corang:
> > @Racer159
> > > IMO the second is more noticeable to a human after the fact. If we wanted to make it more obvious/implement skipping I would prefer to do the skipping GitHub side
> > 
> > 
> > I like this better, I like GitLab where it says passed with warnings and I forgot GitHub doesn't do that, skipping would be much better
> > > Also to the slowdown issue - it would be possible and faster to do this with lookups on OCI tags instead of cloning the repo down.
> > 
> > 
> > This necessitates always publishing the test bundle for every package, should we be doing that?
> 
> It shouldn't req that - removing the spoof would require that but it could follow the skip logic based on package version - IMO if you are changing configuration to the point that it is hard breaking that should be somewhat painful because what you are doing will be painful to delivery and that pain forces the package creator to think about ways to handle that upgrade properly across bundle versions.

The issue I'm remembering was when we changed how postgres secret generation worked for a package, the change the end user to update a variable I think but it was a valid change, however the upgrade test broke because it tried to use the old version of the package with the new version of the bundle

Racer159:
> > > @Racer159
> > > > IMO the second is more noticeable to a human after the fact. If we wanted to make it more obvious/implement skipping I would prefer to do the skipping GitHub side
> > > 
> > > 
> > > I like this better, I like GitLab where it says passed with warnings and I forgot GitHub doesn't do that, skipping would be much better
> > > > Also to the slowdown issue - it would be possible and faster to do this with lookups on OCI tags instead of cloning the repo down.
> > > 
> > > 
> > > This necessitates always publishing the test bundle for every package, should we be doing that?
> > 
> > 
> > It shouldn't req that - removing the spoof would require that but it could follow the skip logic based on package version - IMO if you are changing configuration to the point that it is hard breaking that should be somewhat painful because what you are doing will be painful to delivery and that pain forces the package creator to think about ways to handle that upgrade properly across bundle versions.
> 
> The issue I'm remembering was when we changed how postgres secret generation worked for a package, the change required the end user to update a variable I think but it was a valid change, however the upgrade test broke because it tried to use the old version of the package with the new version of the bundle

For those issues it is possible to phase the configuration through and does make it more obvious that you made a breaking change - this does require some extra thought (see [this example in GitLab](https://github.com/defenseunicorns/uds-package-gitlab/blob/main/bundle/uds-bundle.yaml#L66)) but it also provides examples for how to setup things for delivery to support different versions (in that case it was to support old GitLab with new Valkey in case that configuration was needed).  We can definitely put more thought into this but I think in many cases it can be a good thing to think through before it hits delivery's desk.  It also should probably be split from the general skip part since they are separate problems.

Michael-Kruggel:
For fun, I decided to test `uds-pacakge-gitlab` locally again using `time` on the command and the results were actually different:
My PR:
```
real    16m1.032s
user    5m41.077s
sys     1m55.248s
```
As-is Upgrade Test:
```
real    16m27.494s
user    5m29.581s
sys     2m9.963s
```
It was actually faster this time 🤷‍♂️

zachariahmiller:
> For fun, I decided to test `uds-pacakge-gitlab` locally again using `time` on the command and the results were actually different: My PR:
> 
> ```
> real    16m1.032s
> user    5m41.077s
> sys     1m55.248s
> ```
> 
> As-is Upgrade Test:
> 
> ```
> real    16m27.494s
> user    5m29.581s
> sys     2m9.963s
> ```
> 
> It was actually faster this time 🤷‍♂️

😅  Does that make any sense though? Was the upgrade running between different gitlab versions? All cached images and stuff cleared out? Otherwise its not a representative test.

Racer159:
> > For fun, I decided to test `uds-pacakge-gitlab` locally again using `time` on the command and the results were actually different: My PR:
> > ```
> > real    16m1.032s
> > user    5m41.077s
> > sys     1m55.248s
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > As-is Upgrade Test:
> > ```
> > real    16m27.494s
> > user    5m29.581s
> > sys     2m9.963s
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It was actually faster this time 🤷‍♂️
> 
> 😅 Does that make any sense though? Was the upgrade running between different gitlab versions? All cached images/etc cleared out? Otherwise its not a representative test.

It would be close - the same number of bits would traverse the wire in most cases - but there would be a small slowdown from the tarball creation process (also there would be a small disk hit from more images being in zarf cache).  Slowdown is a small concern but it isn't really my main one (UX and indicators of breakage are more what I am concerned with here)

ericwyles:
Another consideration on rebuilding from source is that the images in the registry could have changed. Rebuilding the same tag from source is not always testing the same package with identical images that are in the released package. We've seen changes like this break app behavior before (even causing apps to not be able to start up).

Michael-Kruggel:
I have made some further changes to this PR to satisfy everyone as much as I can.

1. When cloning the repo, the task will cd into the repo, pull the most recent published tagged package, and create the bundle. I then move that bundle to the expected path and run `setup:k3d-test-cluster` and `deploy:test-bundle` as usual.
2. I added a break flavor only to show how this will behave with my changes. It will be removed before any discussion of merging the branch
3. I left the original behavior of upgrade and renamed it to breaking-change. I did this because then it can be set as allowed to fail and still inform the package maintainer for things that will break for the end user
4. I did everything I knew how to do to try to get some tests to skip and show that, while letting others to actually run the test. With some confidence I can say there is no way to skip individual matrices. It either won't show the skipped status on the PR or it skips all tests if one thing needs to skip the upgrade portion. [EXAMPLE](https://github.com/defenseunicorns/uds-common/actions/runs/10310424456?pr=206)

Michael-Kruggel:
@zachariahmiller @Racer159 @mjnagel @ericwyles 

I have made some further changes to this PR to satisfy everyone as much as I can.

1. When cloning the repo, the task will cd into the repo, pull the most recent published tagged package, and create the bundle. I then move that bundle to the expected path and run `setup:k3d-test-cluster` and `deploy:test-bundle` as usual.
2. I added a break flavor only to show how this will behave with my changes. It will be removed before any discussion of merging the branch
3. I left the original behavior of upgrade and renamed it to breaking-change. I did this because then it can be set as allowed to fail and still inform the package maintainer for things that will break for the end user
4. I did everything I knew how to do to try to get some tests to skip and show that, while letting others to actually run the test. With some confidence I can say there is no way to skip individual matrices. It either won't show the skipped status on the PR or it skips all tests if one thing needs to skip the upgrade portion. [EXAMPLE](https://github.com/defenseunicorns/uds-common/actions/runs/10310424456?pr=206)

tasks.yaml
```diff
@@ -89,7 +90,12 @@ tasks:
         with:
           assessment_results: /tmp/oscal-assessment-results.yaml
 
-  - name: test-upgrade
+  - name: test-latest-tag-flavor
```
Racer159, line 93: we shouldn't call tasks the same name across files since that can make it more confusing to know what is called where.

tasks/upgrade.yaml
```diff
@@ -0,0 +1,95 @@
...
+            fi
+
+            cd ..
+
+            ./uds run test-upgrade --no-progress --set FLAVOR=${FLAVOR} ${{ .inputs.options }} --no-progress
+
+          # If flavor doesn't exist
+          else
+            cd ..
+            echo "::warning::⚠️ Latest tag does not have flavor:${FLAVOR}"
```
Racer159, line 92: This is possible to show in GH with callable workflows: https://github.com/defenseunicorns/uds-common/pull/211

tasks.yaml
Racer159: (also pretty sure this is dead code)

tasks/upgrade.yaml
Racer159: Doing so would also avoid the case where this task fails because the repo is new and has no tags.

.github/workflows/test-deploy.yaml
```diff
@@ -37,17 +37,15 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 20
     strategy:
+      fail-fast: false
       matrix:
-        flavor: [test]
-        type: [install, upgrade]
+        flavor: [test, break]
+        type: [install, upgrade, breaking-change]
```
Racer159, line 43: this feels like it adds a lot to these matrix runs (runner concurrency starts to be a question on the repos that need the bigger runners).  I would say either we care about the breaking change case in CI and that is the one that runs or we don't and the other is the one that runs.

Racer159 (changes requested):

.github/workflows/test-deploy.yaml
Michael-Kruggel: Removed breaking-change

Michael-Kruggel:
This now behaves like requested. Example of a skip for a non-existing flavor in tagged repo is here: [EXAMPLE](https://github.com/defenseunicorns/uds-common/actions/runs/10328532900?pr=206)

Racer159 (approved):

PR #222, opened 2024-08-29
feat: uds-security-hub scan common action

partkyle:
This is an action that is meant to be used for scanning a zarf package that is created by a workflow.

[Here is an example repo that is using it](https://github.com/defenseunicorns/uds-security-hub-common-action-test/blob/main/.github/workflows/scan.yaml).


.github/actions/security-hub-scan/action.yaml
```diff
@@ -0,0 +1,68 @@
...
+        tmpdir=$(mktemp -d)
+        echo "dir=$tmpdir" >> $GITHUB_OUTPUT
+      shell: bash
+
+    - name: use uds-security-hub to scan a package
+      id: scan
+      run: |
+        OUTPUT_FILE=${{ steps.tmpdir.outputs.dir }}/cve-scan.${{ inputs.format }}
+
+        docker run --rm \
```
zachariahmiller, line 41: I'm not opposed to this, but are there any references to why this is a script in an action calling a docker container rather than a uds task. Also why this functionality couldnt be vendored into uds-cli so it would just be `uds scan <thing to scan>` to execute?

.github/actions/security-hub-scan/action.yaml
```diff
@@ -0,0 +1,68 @@
+name: security-hub-scan
```
zachariahmiller, line 1: Common pattern we have is that the action would be tested in this repo, so if you wouldnt mind implementing this getting called as part of the workflow/workflows (test/publish) that would be great.

zachariahmiller (changes requested):

.github/actions/security-hub-scan/action.yaml
partkyle: ok, I see other examples that have that in this repo that I can follow.

this is going to need a zarf package for a test, and ideas on how to support that?

zachariahmiller: a zarf package is built and published as part of this repo to provide an example and facilitate the repo's CI being able to test itself. It is currently [podinfo](https://github.com/defenseunicorns/uds-common/pkgs/container/packages%2Fuds%2Fpodinfo) and in the [next release](https://github.com/defenseunicorns/uds-common/pull/218) will be updated to nginx to provide a more representative example

.github/actions/security-hub-scan/action.yaml
partkyle: This is an agreed upon interim solution, but I do think what you mention makes a lot of sense long term.

Racer159: Linking this to the most relevant issue I think is there to give folks breadcrumbs: https://github.com/defenseunicorns/uds-security-hub/issues/176 cc @koesbong as well if there is more context to give (I loosely followed some stuff but only have a slice myself)

.github/workflows/test.yaml
```diff
@@ -55,3 +55,6 @@ jobs:
       flavor: ${{ matrix.flavor }}
       type: ${{ matrix.type }}
     secrets: inherit # Inherits all secrets from the parent workflow.
+
+  test-security-hub-scan:
```
Racer159, line 59: just a note on having this in the end to end test workflow - this is probably fine for now but it may be beneficial to think about splitting it - especially if you wanted to use in-package SBOMs in the future (the tests currently exclude those to save time and SBOMs are only generated on release) - the can results are somewhat useful on PRs for tracking though I would argue would be more so for customers on release and/or in a cron job workflow that runs more consistently (which could be a separate workflow that specifies a timeframe to run within)

partkyle: I added this here because I only saw this one place for testing, but I'd be happy to split that out if that's preferable.

.github/actions/security-hub-scan/action.yaml
partkyle: is there anything I should do here? Should I take a stab at making this work with the tasks system, even though it's currently going to be running with the docker command?

.github/actions/security-hub-scan/action.yaml
partkyle: I was able to use that local package to test this workflow.

partkyle:
@zachariahmiller @Racer159 Can you both take a look at this and see if I have addressed the above comments? Thanks!

.github/actions/security-hub-scan/action.yaml
zachariahmiller: Yeah I think it still probably makes sense to make this a task and call it in the action or from workflow. We do the same where it's wrapped in a docker run for our playwright tests. 

That way someone can run locally via `uds run` as well.

partkyle: @zachariahmiller I took a stab at writing this as a uds task. I left the cosign bit out because it requires cosign to be installed, so the cosign portion is only in the action at the moment. Can you take another look when you get a chance?

partkyle:
@Racer159 @zachariahmiller any more feedback on this PR for things that need changed?

zachariahmiller:
> @Racer159 @zachariahmiller any more feedback on this PR for things that need changed?

@partkyle I'm out this week so I will defer to @Racer159  on this. Thanks.

.github/workflows/test-security-hub-scan.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+      # run setup to get uds to use zarf
+      - name: Environment setup
+        uses: ./.github/actions/setup
+
+      - name: build example package
+        id: setup
+        run: |
+          uds run create-nginx-package --no-progress
+          echo "path="$(ls *.zst) >> "$GITHUB_OUTPUT"
```
Racer159, line 27: we should avoid wildcarding - this will not pickup application repos that have multiple packages properly (i.e. mattermost which has the main app and a plugins package)

.github/workflows/test-security-hub-scan.yaml
```diff
@@ -0,0 +1,47 @@
...
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+
+      # run setup to get uds to use zarf
+      - name: Environment setup
+        uses: ./.github/actions/setup
+
+      - name: build example package
+        id: setup
+        run: |
+          uds run create-nginx-package --no-progress
```
Racer159, line 26: we should not rely on app-specific tasks in callable workflows (other apps should be able to use this flow as is without needing to add an extra `create-nginx-package` task)

.github/workflows/test-security-hub-scan.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+      - name: run scan
+        id: scan
+        uses: ./.github/actions/security-hub-scan
+        with:
+          path: ${{ steps.setup.outputs.path }}
+
+      # verify that the outputs and the cosign bundle are correct
+      - name: install cosign
+        uses: sigstore/cosign-installer@4959ce089c160fddf62f7b42464195ba1a56d382 # v3.6.0
```
Racer159, line 38: given this takes 2s it should likely be simplified to happen in the env setup (we will want to use cosign for package / bundle signing later too)

.github/workflows/test.yaml
Racer159: Thinking on this more I do think it would be good to split out into it's own workflow - this also does not matrix flavors so we will only get one result from it.

tasks/security-hub.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - name: PKG_PATH
+    description: Path to the zarf package to scan
+  - name: OUTPUT_FILE
+    description: Path to write the results into
+  - name: OUTPUT_FORMAT
+    description: Scan output format
+    default: json
+  - name: VERSION
+    description: Version of uds-security-hub to use
+    default: v0.1.0
```
Racer159, line 11: Variables of similar names can easily get mangled - `VERSION` for example may override the default of something else called `VERSION` elsewhere in Maru if the tasks are composed - these should be either refactored into `inputs` (if they are unlikely to change often) or be namespaced with something like `SECURITY_HUB_`.

.github/actions/security-hub-scan/action.yaml
```diff
@@ -0,0 +1,59 @@
...
+          --set PKG_PATH=${{ inputs.path }} \
+          --set OUTPUT_FILE=${OUTPUT_FILE} \
+          --set OUTPUT_FORMAT=${{ inputs.format }} \
+          --set VERSION=${{ inputs.version }}
+
+        echo "cve-scan=${OUTPUT_FILE}" >> "$GITHUB_OUTPUT"
+      shell: bash
+
+    - name: install cosign
+      uses: sigstore/cosign-installer@4959ce089c160fddf62f7b42464195ba1a56d382 # v3.6.0
```
Racer159, line 40: We can likely assume this happens in env setup per the other comment

.github/workflows/test-security-hub-scan.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+      - name: verify cosign results
+        env:
+          COSIGN_EXPERIMENTAL: "true"
+        run: |
+          cosign verify-blob --bundle ${{ steps.scan.outputs.cosign-bundle }} ${{ steps.scan.outputs.cve-scan }} \
+              --certificate-identity-regexp="https://github.com/${{ github.repository }}/" \
+              --certificate-oidc-issuer="https://token.actions.githubusercontent.com"
+        shell: bash
```
Racer159, line 47: Is there a reason to double verify?

Racer159 (changes requested):

.github/workflows/test-security-hub-scan.yaml
partkyle: I have changed it to a pull request triggered workflow, so it should be fine to use this task, correct?

.github/workflows/test-security-hub-scan.yaml
partkyle: added to the setup step

tasks/security-hub.yaml
partkyle: update the variable name

.github/actions/security-hub-scan/action.yaml
partkyle: done

.github/workflows/test-security-hub-scan.yaml
partkyle: no, this is not needed because the step verifies it already, so that will be sufficient for a test

.github/workflows/test-security-hub-scan.yaml
Racer159: That should be fine for now - eventually we will fix Maru and then this can be callable again with a reference to `create:package`

.github/actions/security-hub-scan/action.yaml
partkyle: added tests, so I am resolving this

.github/workflows/test-security-hub-scan.yaml
partkyle: @Racer159 Ok, I think I have this resolved per your comments. Can you take a look?

PR #224, opened 2024-09-05
fix: mandate yamllint at least 1.30.0 to guarantee features in use

JoeHCQ1:
Version 1.29.0 introduced something with the ignore filepaths. If using v1.28.0 or earlier you get the error `invalid config: ignore should contain file patterns`.

Anchors, in use in at least the uds-package-mattermost repo are not recognized in v1.29.0, hence setting v1.30.0 as the minimum allowed version.

As the code stood, having say version 1.26.0 installed satisfied the `pip install yamllint` command but `uds run lint:yaml` failed frustrating this end user.

### DIFF
--- a/tasks/lint.yaml
+++ b/tasks/lint.yaml
@@ -7,7 +7,7 @@ tasks:
     description: Install linting tool dependencies
     actions:
       - description: Install yamllint via pip
-        cmd: CMD=pip && which $CMD || CMD=pip3 && $CMD install yamllint
+        cmd: CMD=pip && which $CMD || CMD=pip3 && $CMD install yamllint>=1.30.0
 
   - name: yaml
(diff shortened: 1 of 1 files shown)


JoeHCQ1:
I tested this by trying to run the `yaml:lint ` job in the `uds-package-mattermost` repo on pinned yamllint versions until it passed. Note, I've encountered this problem on other repos too, I just took the time to fix it for the first time today, and happened to be in the mattermost repo.

mjnagel (approved):
LGTM - good catch on this one.

JoeHCQ1:
As a further point of testing to guard against a syntax error,  I changed my lint import to import this commit, and then ran `uds run lint:deps` `uds run lint:yaml` and both worked.

PR #225, opened 2024-09-05
chore: add uds-marketplace to codeowners

andrewg-xyz:


daveworth (approved):

mjnagel (approved):

mjnagel (dismissed):

.github/workflows/test-deploy.yaml
```diff
@@ -12,6 +12,23 @@ on:
...
+      - docs/**
+      - .gitignore
+      - renovate.json
+      - .release-please-config.json
+      - release-please-config.json
+      - CODEOWNERS
+      - LICENSE
+      - CONTRIBUTING.md
+      - SECURITY.md
+      - config/renovate.json5
```
mjnagel, line 31: I think we'll need to update the docs-shim workflow, specifically the [name of each job here](https://github.com/defenseunicorns/uds-common/blob/update-codeowners/.github/workflows/ci-docs-shim.yaml#L26). If set to `test-deploy (${{ matrix.type }}, ${{ matrix.flavor }}) / deploy` I think that would work?

mjnagel: (I think the paths-ignore changes here can be reverted as well since they are already present on the parent workflow here: https://github.com/defenseunicorns/uds-common/blob/main/.github/workflows/test.yaml#L8)

daveworth (approved):

mjnagel (approved):

PR #235, opened 2024-09-19
feat: add options support to setup to set vars and other flags on uds-core

Racer159:


brandtkeller (approved):
Each task with use of the options contains the defined `options` key under associated `inputs` field.

andrewg-xyz (approved):

PR #238, opened 2024-09-20
chore: modify helm matches to handle git and helm

mjnagel:
Current matches will greedily match git charts in the helm repo match, causing lookup failure warnings. This regex changes to detect datasource based on presence of `.git`.


Racer159 (approved):
lgtm

PR #241, opened 2024-09-23
fix: allow dependent bundle commands to be run on upgrade tests

Racer159:
This adds an input to allow dependency commands to run in the cloned repo prior to bundle create as needed in some repos (i.e. something needing `uds run dependencies:create`)

### DIFF
--- a/tasks/upgrade.yaml
+++ b/tasks/upgrade.yaml
@@ -32,6 +32,10 @@ tasks:
         description: The architecture of the package to create
         default: ${UDS_ARCH}
 
+      dep_commands:
+        description: Additional commands to run inside the old repo to setup dependencies
+        default: ""
+
     actions:
       - task: utils:determine-repo
         with:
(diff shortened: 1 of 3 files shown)


andrewg-xyz (approved):

tasks/upgrade.yaml
```diff
@@ -32,6 +32,10 @@ tasks:
         description: The architecture of the package to create
         default: ${UDS_ARCH}
 
+      dep_commands:
+        description: Additional commands to run inside the old repo to setup dependencies
```
andrewg-xyz, line 36: I may be missing context, but what is meant "run inside the old repo..."

ericwyles (approved):

tasks/upgrade.yaml
Racer159: The upgrade task here clones in the older repo at the last tag to make the bundle to handle version updates more dynamically (useful if chart names are added or packages change in more significant ways that are incompatible) - that's what all the git stuff is doing toward the top of the task.

PR #254, opened 2024-09-27
feat!: update uds common gh actions to use uds run conditionals

zachariahmiller:
This issue convert all the common actions into tasks and updates the workflows to all be functioning reusable (callable) workflows.

To upgrade to this version of uds-common once it is released, package developers will need to replace their existing workflows with new ones that call the callable workflows. This is being done by example already in this pr for uds-common itself, but will require slight modification in the package repos.

The specific deviations the package developer will need to make are to update the two uses blocks to be remote refs.
Specifically: `uses: ./.github/workflows/callable-test.yaml  `would become `uses: defenseunicorns/uds-common/.github/workflows/callable-test.yaml@sha`
and `uses: ./.github/actions/test-flavor`  would become `uses: defenseunicorns/uds-common/.github/actions/test-flavor@sha`
Similar updates will need to happen to the lint, release and doc shim workflows in each package repo. After making these changes the branch protections rules will need to be updated to match the new naming schemes.

This update will also require UDS CLI version >= 0.17.0

PR #258, opened 2024-10-03
chore!: add shell linting to uds-common linting

Racer159:
This adds shellcheck for linting shell scripts in UDS Packages.

> [!CAUTION]
> :warning: **BREAKING CHANGE** - shellcheck is on by default and will now fail linting if a package has scripts that do not comply.


tasks/lint.yaml
```diff
@@ -20,4 +27,93 @@ tasks:
     description: Run linting checks on OSCAL
     actions:
       - description: Lint OSCAL files
-        cmd: lula tools lint -f $OSCALFILES
+        cmd: lula tools lint -f "$OSCALFILES"
```
andrewg-xyz, line 30: iirc this breaky if the '$OSCALFILES' don't exist

Racer159: I can add an if on it - _and_ now that we have this https://github.com/defenseunicorns/uds-common/blob/main/docs/package_integration/oscal-guidelines.md I'll update the default too.

Racer159: added but @brandtkeller thoughts on including this check in Lula directly? - also maybe not bc I think eventually we'd want to force this on you...

Racer159: added a warning for now - we may eventually want to force this on people.

Racer159: (also this is broken until https://github.com/defenseunicorns/uds-common/pull/254 pulls in an update to uds-cli)

Racer159:
(scan lint is blocked until uds cli update comes through https://github.com/defenseunicorns/uds-common/pull/254)

andrewg-xyz (approved):

zachariahmiller (dismissed):

zachariahmiller (approved):

PR #262, opened 2024-10-04
feat(task)!: add optional config input to create, deploy package and bundle tasks

justinthelaw:
Adds an optional `config` input for the `[create | deploy]:[test-bundle | package]` task so that end-users can supply a `[zarf | bundle]-config.yaml` for create-time and deploy-time configuration.

## BREAKING CHANGE

- `[create | deploy]:test-bundle` now requires you to specify the full-path to the `UDS_CONFIG` (e.g. `bundle/uds-config.yaml`) that you intend on using for a UDS bundle creation and deployment. The path to the `UDS_CONFIG` is no longer assumed to be the same path as the UDS bundle's location.  If you are accepting the defaults however no change is necessary.


justinthelaw:
Example down-stream usage: https://github.com/defenseunicorns/leapfrogai/blob/badcabfc54bd2eab352920d1f247ba59d3963610/tasks/create.yaml#L130

Need for configuration such as in LeapfrogAI's vLLM: https://github.com/defenseunicorns/leapfrogai/tree/736-feat-leapfrogai-ui-registry1-flavor/packages/vllm

justinthelaw:
Additional context from Slack:

> LeapfrogAI is a mono-repo with many sub-packages, so we have to specify the ZARF_CONFIG when running UDS tasks form the root of the repository. Additionally, the ZARF_CONFIG for vLLM is a declarative way to define what model files you want to be downloaded and Zarf-injected into the cluster, and also additional engine configurations at deploy-time. It allows delivery/mission-success engineers to maintain multiple Zarf configurations based on what models and parameters they want to deploy.
> Example zarf-config.yaml: https://github.com/defenseunicorns/leapfrogai/blob/main/packages/vllm/zarf-config.yaml

corang (approved):

tasks/create.yaml
```diff
@@ -14,8 +14,14 @@ tasks:
...
         description: The architecture of the package to create
         default: ${UDS_ARCH}
+      config:
+        description: Optionally supply a path to the desired Zarf Configuration for the package
+        default: ""
+        required: false
     actions:
-      - cmd: |
+      - env:
+          - ZARF_CONFIG=${{ .inputs.config }}
```
Racer159, line 23: Is LFAI standardized enough to match the UDS Config when dealing with bundles where it also takes `path`? `UDS_CONFIG=${{ .inputs.path }}/${{ .inputs.config }}`

Racer159 (changes requested):

tasks/create.yaml
justinthelaw: I standardized the Zarf deploy/create tasks with the UDS ones. Both now use the original path and the config filename. Both declare the `[ZARF | UDS]_CONFIG` via the `env` field instead of in-line of the `cmd`. How does that look to you now?

tasks/deploy.yaml
```diff
@@ -6,6 +6,10 @@ tasks:
       path:
         description: Path relative to the repositories root where the zarf.yaml and .tar.zst lives
         default: .
+      config:
+        description: Optionally provde a path to the package config file
```
Racer159, line 10:
```suggestion
        description: Optionally provide a path to a Zarf config file
```

tasks/create.yaml
```diff
@@ -14,8 +14,14 @@ tasks:
       architecture:
         description: The architecture of the package to create
         default: ${UDS_ARCH}
+      config:
+        description: Optionally provde a path to the package config file
```
Racer159, line 18:
```suggestion
        description: Optionally provide a path to a Zarf config file
```

Racer159 (changes requested):

Racer159 (approved):

zachariahmiller (approved):

PR #267, opened 2024-10-07
fix!: update the publish task to avoid name collision

Racer159:
## Description

This updates the publish task to avoid name collisions and cleans up the main tasks file more.

## Checklist before merging

- [X] Tests run, docs added or updated as needed


andrewg-xyz (approved):

tasks.yaml
```diff
@@ -81,16 +47,34 @@ tasks:
...
+      - description: Create the test bundle
+        task: create:test-bundle
+
+      - description: Deploy the test bundle
+        task: deploy:test-bundle
+
+      - description: Test the bundle
+        task: test
+
+# CI will execute the following (via .github/actions/test and .github/actions/release) so they need to be here with these names
```
zachariahmiller, line 62:
```suggestion
# CI will execute the following (via .github/workflows/callable-test.yaml and .github/workflows/callable-publish.yaml) so they need to be here with these names
```

zachariahmiller (approved):

PR #283, opened 2024-10-09
feat: support gitlab

zachariahmiller:
## Description

- break out components of environment-setup
- add gitlab registry login 
- add generic registry login task
- add gitlab ci components in `templates`
- add example .gitlab-ci.yaml that builds uds-common on gitlab

Fixes: https://github.com/defenseunicorns/uds-software-factory/issues/92

## Checklist before merging


templates/publish.yml
```diff
@@ -0,0 +1,118 @@
...
+        if ! uds-releaser check ${FLAVOR}; then echo "No release necessary"; exit 0; fi
+      fi
+
+    # Modify zarf package and uds bundle to have the correct version
+    - |
+      if [[ "$DRY_RUN" == "true" ]]; then
+        echo "Dry run, skipping release command: uds-releaser update-yaml ${FLAVOR}"
+      else
+        uds-releaser update-yaml ${FLAVOR}
+      fi
```
Racer159, line 65: if it is dry-running the publish wouldn't we still want to run these? They should run relatively quick (and it would be less logic to maintain)

templates/publish.yml
```diff
@@ -0,0 +1,118 @@
...
+          uds run $TASK_NAME --set FLAVOR="$[[ inputs.flavor ]]"\
+                                     --set VERSION="$(uds zarf tools yq ".flavors[] | select(.name == \"${FLAVOR}\") | .version" releaser.yaml)"\
+                                     --set TARGET_REPO="$[[ inputs.targetRepo ]]"\
+                                     ${OPTIONS}
+        else
+          uds run $TASK_NAME --set FLAVOR="$[[ inputs.flavor ]]"\
+                                     --set VERSION="$(uds zarf tools yq ".flavors[] | select(.name == \"${FLAVOR}\") | .version" releaser.yaml)"\
+                                     --set TEAM="$[[ inputs.team ]]"\
+                                     ${OPTIONS}
+        fi
```
Racer159, line 85: rather than copying all of this we might be able to simplify into something like how dry-run and options interact

templates/test.yml
```diff
@@ -0,0 +1,89 @@
...
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+spec:
+  inputs:
+    runsOn:
+    flavor:
+    type:
+    artifact_dir:
+      default: .ci_artifacts
+    reports-path:
```
Racer159, line 11: same comment on input casing

templates/test.yml
```diff
@@ -0,0 +1,89 @@
...
+      when: always
+  allow_failure:
+    exit_codes: 222 # using something here to exit with a warning
+  script:
+    - |
+      # Check if any tag exists
+      TAG="$(git describe --tags "$(git rev-list --tags --max-count=1)" 2>/dev/null || echo '')"
+
+      if [ -z "$TAG" ]; then
+        echo "No Git tags found. Skipping step."
```
Racer159, line 35:
```suggestion
        echo "No Git tags found. Skipping upgrade check."
```

Racer159 (changes requested):

templates/publish.yml
ericwyles: I think we can keep the `update-yaml` but the check flavor actually stops the processing if there is "No release necessary" so it actually keeps any of the other dry runs from happening. I'll test what happens if we keep `update-yaml` in there.

tasks/publish.yaml
```diff
@@ -56,6 +64,7 @@ tasks:
         default: ${BUNDLE_NAME}
     actions:
       - task: utils:determine-repo
+        if: ${{ eq .variables.TARGET_REPO "" }}
```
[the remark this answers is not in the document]
ericwyles: I put this in here to bypass the utils:determine-repo completely. The idea was if you need to completely set the TARGET_REPO to something else from the specific project, you can set it and what you set will be used instead. I put this in while we were testing publish by posting to GitLab's registry rather than put more github vs gitlab specific stuff inside determine-repo because the inputs themselves are quite a bit different.

You can see this used in `templates/publish.yml` if the person wiring up the job has a specific target repo they want to use. If we are just going to GHCR for everything right now I can take this (and the logic in `templates\publish.yml`) out completely as it won't really be used yet.

Thoughts on that?

templates/publish.yml
ericwyles: Actually I can refactor the `check flavor` to still run but not exit if dry run is enabled.

templates/publish.yml
ericwyles: I will do this but want to resolve the `TARGET_REPO` conversation above first because I might just delete that.

templates/publish.yml
```diff
@@ -0,0 +1,118 @@
+# Copyright 2024 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+spec:
+  inputs:
+    runsOn:
+    flavor:
+    artifact_dir:
```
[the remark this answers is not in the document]
ericwyles: Fixed!

templates/test.yml
ericwyles: Fixed.

tasks/publish.yaml
```diff
@@ -7,6 +7,13 @@ includes:
 variables:
   - name: FLAVOR
     default: upstream
+  - name: TARGET_REPO
+  - name: VERSION
+    # x-release-please-start-version
+    default: 1.1.2
+    # x-release-please-end
```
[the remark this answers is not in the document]
ericwyles: Fixed this

tasks/publish.yaml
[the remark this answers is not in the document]
ericwyles: Thought about it and left the `TARGET_REPO` override in here, but moved the logic that checks it down into `utils:determine-repo` so it's all contained in there. You can still `--set` it on the command line and it works.

templates/publish.yml
ericwyles: Refactored this so that the `uds-releaser check` still runs all the time but if dry run is on it doesn't `exit` so that other commands can actually still execute with the dry run flag.

Updated `uds-releaser update-yaml` to run all the time as-is because it is harmless.

templates/publish.yml
ericwyles: Fixed it up!

templates/test.yml
ericwyles: Ended up refactoring inputs in general to be dashes instead of camelCase throughout the GitLab CI components so it's all consistent.

releaser.yaml
```diff
@@ -0,0 +1,10 @@
+# Copyright 2024 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+flavors:
+  - name: upstream
+    version: 1.1.2
```
corang, line 6: Given us still using release-please we should add the comment annotations around all of these versions right?

tasks/publish.yaml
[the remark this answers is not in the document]
Racer159: yep that is more what I meant - will review further!

templates/README.md
```diff
@@ -0,0 +1,53 @@
...
+1. Test
+1. Publish
+1. Scorecard
+
+The .gitlab-ci.yml at the root of this repo serves as an example and test of how to call these components. That being said, a general example of how to include these components in a GitLab `uds-package` is as follows:
+
+```yaml
+include:
+  # include the component located in the current project from the current SHA
+  - component: $CI_SERVER_FQDN/$CI_PROJECT_PATH/commitlint@$CI_COMMIT_SHA
```
corang, line 16: If this is example usage for uds-packages in gitlab then isn't `CI_COMMIT_SHA` incorrect? They should be referring to a specific tag right?

templates/publish.yml
```diff
@@ -0,0 +1,110 @@
...
+    # Publish Package
+    - |
+        if uds run --list | grep -q 'publish-package'; then
+          TASK_NAME="publish-package"
+        else
+          TASK_NAME="publish-release"
+        fi
+
+        ARGS="--set FLAVOR=\"$[[ inputs.flavor ]]\" \
+              --set VERSION=\"$(uds zarf tools yq ".flavors[] | select(.name == \"${FLAVOR}\") | .version" releaser.yaml)\""
```
corang, line 69:
```suggestion
              --set VERSION=\"$(uds-releaser show ${FLAVOR} --version-only)\""
```

corang (changes requested):

Racer159 (approved):

releaser.yaml
Racer159: yes and edit the release please config to include this file

templates/README.md
Racer159: yes

releaser.yaml
ericwyles: Done!

Racer159 (approved):

corang (approved):

PR #284, opened 2024-10-09
chore: disable stale PR rebasing for renovate

mjnagel:
## Description

We currently have a confusing combination of renovate options:
- `"rebaseWhen": "conflicted",`: Only rebases if a PR has conflicts
- `"rebaseStalePrs": true`: Sets `rebaseWhen` to `behind-base-branch` overriding ^

Imo it makes more sense to be less aggressive with rebasing in the base config and allow individual repos to override this. If this PR is not accepted as it stands we should still resolve the confusion around these two options (align them).

## Checklist before merging


Racer159 (approved):
we can try this sure

PR #303, opened 2024-10-15
chore: adds check for addlicense

UncleGedd:
## Description

Adds a check for the `addlicense` binary before running the `license` and `fix-license` tasks


andrewg-xyz (dismissed):

UncleGedd:
My bad, meant to bail if `addlicense` is not installed

Racer159 (approved):

PR #324, opened 2024-10-25
fix: pull mv command to use latest version

Racer159:
## Description

...

## Checklist before merging


Racer159:
bundle change tested here: https://github.com/defenseunicorns/uds-software-factory/pull/101

ericwyles (approved):
I had to hunt for what {$VAR%-*} does lol

PR #329, opened 2024-10-30
fix: gitlab publish being broken due to bash weirdness

corang:
## Description

When you have escaped double quotes in a bash variable bash does some weird single quote things that the command actually ran looks like this
```
--set 'FLAVOR="upstream"'
```
which breaks our stuff
Fix is to switch to a bash array for keeping track of everything:
```
        ARGS=(
          --set FLAVOR="$[[ inputs.flavor ]]"
          --set VERSION="$(uds-releaser show $[[ inputs.flavor ]] --version-only)"
        )

          ARGS+=(--set TARGET_REPO="$[[ inputs.target-repo ]]")
```

## Checklist before merging

### DIFF
--- a/templates/publish.yml
+++ b/templates/publish.yml
@@ -65,16 +65,20 @@ publish:
           TASK_NAME="publish-release"
         fi
 
-        ARGS="--set FLAVOR=\"$[[ inputs.flavor ]]\" \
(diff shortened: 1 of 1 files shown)


corang:
successful gitlab ci run with it https://gitlab.uds.is/defenseunicorns/uds/packages/uds-package-juice-shop/-/jobs/236#L110

ericwyles (approved):

PR #332, opened 2024-11-04
feat: allow keycloak_group to be specified as input in create-doug-user task

JeffResc:
## Description
Allow `keycloak_group` input in `create-doug-user` task to add doug user to specified group, for example:

```yaml
- name: create-admin-doug
  actions:
    - task: setup:create-doug-user
      with:
        keycloak_group: /UDS Core/Admin
```

## Checklist before merging

### DIFF
--- a/tasks/setup.yaml
+++ b/tasks/setup.yaml
@@ -54,7 +54,7 @@ tasks:
     inputs:
       keycloak_group:
         description: Group to add user to
-        default: ""
+        default: "$KEYCLOAK_GROUP"
         required: false
     actions:
       - description: Creating the 'doug' user in the 'uds' realm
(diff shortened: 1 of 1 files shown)


tasks/setup.yaml
```diff
@@ -51,9 +51,15 @@ tasks:
 
   - name: create-doug-user
     description: Creates a user named 'doug' in the uds realm of keycloak (using the default admin account)
+    inputs:
+      keycloak_group:
+        description: Group to add user to
+        default: ""
```
Racer159, line 57:
```suggestion
        default: "$KEYCLOAK_GROUP"
```

Racer159: we should maintain variable compatibility - as is this will always ignore the variable at the top

Racer159 (changes requested):

tasks/setup.yaml
JeffResc: nice catch!

Racer159 (approved):

PR #353, opened 2024-12-05
feat: grype scan custom action

partkyle:
## Description

A reusable workflow for scanning all of the container images in a zarf.yaml file.

## Checklist before merging


zachariahmiller:
A couple thoughts:

1. I dont think this belongs in a dedicated callable workflow. In the future maybe there is a reason to add a "scan" workflow, but for now it should probably be part of lint.
2. We should not rely on a github action to install grype, but rather a task. The grype install process (not using the action) should be added under tasks/actions.yaml 
3. This needs to be updated to support GitLab as well, which all the other common functions now do. The gitlab components are implemented under `templates/`
4. I think its fine in a PR workflow to scan base on the zarf.yaml, which is the use case here. Just to note: in terms of anything speaking to posture of an actual published though it should happen from images actually present in the artifact, not based on the zarf.yaml as image tags arent necessarily immutable.

partkyle:
> 2. We should not rely on a github action to install grype, but rather a task. The grype install process (not using the action) should be added under tasks/actions.yaml 

I addressed this and the setup is in the actions.yaml file and has been removed from the github action.

partkyle:
This came up in response to a slack message, but we have decided on a different approach, so I'm closing this issue.

PR #367, opened 2024-12-13
feat: move mirror ci to publish step

corang:
## Description

...

## Checklist before merging

### DIFF
--- a/templates/publish.yml
+++ b/templates/publish.yml
@@ -90,7 +90,6 @@ publish:
 
     # Mirror package to github
     - |
-        set -x
         if [[ "$[[ inputs.mirror ]]" == "true" ]]; then
           export PKG_VERSION="$(uds-releaser show $[[ inputs.flavor ]] --version-only)"
           export PKG_NAME="$(yq '.metadata.name' zarf.yaml)"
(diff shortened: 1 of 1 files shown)


templates/publish.yml
```diff
@@ -84,6 +86,23 @@ publish:
...
+          export PKG_NAME="$(yq '.metadata.name' zarf.yaml)"
+          curl -sS --request POST \
+            --form "token=${CI_JOB_TOKEN}" \
+            --form "variables[VERSION]=${PKG_VERSION}" \
+            --form "variables[PROJECT_NAME]=${CI_PROJECT_NAME}" \
+            --form "variables[PROJECT_NAMESPACE]=${CI_PROJECT_NAMESPACE}" \
+            --form "variables[PACKAGE_NAME]=${PKG_NAME}" \
+            --form "variables[FLAVOR]=$[[ inputs.flavor ]]" \
+            --form "variables[ARCH]=$[[ inputs.arch ]]" \
+            --form ref=main "${CI_API_V4_URL}/projects/11/trigger/pipeline"
```
Racer159, line 102: This needs to be configurable if it is to live in common

Racer159 (changes requested):

templates/publish.yml
```diff
@@ -19,6 +19,10 @@ spec:
       default: ""
     team:
       default: uds
+    mirror:
+      default: "true"
+    mirror-project-id:
+      default: "11"
```
Racer159, line 25:
```suggestion
```

IMO we should not have this default here since it is env specific and not really a default for all envs.

Racer159 (changes requested):

Racer159 (approved):

PR #384, opened 2025-01-16
docs: added link to search chainguard for images.

JoeHCQ1:


docs/uds-packages/guide.md
```diff
@@ -8,7 +8,7 @@ Integrating a Package fundamentally means:
 1. Creating a repository `uds-package-<name>` from [uds-package-template](https://github.com/defenseunicorns/uds-package-template)
 2. Integrating the upstream helm chart as a zarf package `zarf.yaml` to build a declarative OCI artifact
 3. Adding a UDS package Custom Resource `uds-package.yaml` to integrate with UDS Core via Pepr
-4. Build a 'zero CVE' package by replacing images with a `*-unicorn` flavored image
+4. Build a 'zero CVE' package by replacing images with a `*-unicorn` flavored image. Search for chainguard images here: <https://images.chainguard.dev/?>.
```
zachariahmiller, line 11: I dont think we want to actually make this reference explicit here as while currently that means this image source unicorn flavor doesnt necessarily mean and wont necessarily mean this image source in the future. I dont actually feel particularly strong about this though. I'd defer to @andrewg-xyz if he has thoughts.

mjnagel: Agreed - I would say this is primarily for us internally since we don't always have a cgr image option but doesn't necessarily belong in this doc for now. Going to close this PR - if anyone wants to object/revisit feel free to reopen.

PR #385, opened 2025-01-16
feat: add ability to specify base repo when publishing

corang:


tasks/publish.yaml
```diff
@@ -98,6 +100,7 @@ tasks:
       - task: utils:determine-repo
         with:
           team: ${{ .variables.TEAM }}
+          base_repo: ${{ .variables.BASE_REPO }}
```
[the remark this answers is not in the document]
zachariahmiller: I think it also needs to be added to the two pull tasks

tasks/publish.yaml
```diff
@@ -70,6 +71,7 @@ tasks:
       - task: utils:determine-repo
```
zachariahmiller, line 71: Can you refactor utils:determine repo so that TARGET_REPO is only the output and not a possible input the task operates conditionally based on?

zachariahmiller (changes requested):

tasks/utils.yaml
```diff
@@ -23,16 +22,11 @@ tasks:
...
-        # if the caller explicitly set a value for TARGET_REPO we'll use it
-        if: ${{ ne .variables.TARGET_REPO "" }}
-        cmd: echo "Using supplied value TARGET_REPO=${TARGET_REPO}"
       - description: Determine repository for the given flavor/type of release
-        if: ${{ eq .variables.TARGET_REPO "" }}
         cmd: |
           repo="${{.inputs.base_repo}}"
           # unicorn flavor = private repository
-          if [ "${{.inputs.flavor}}" = "unicorn" ]; then
+          if [ "${{.inputs.flavor}}" = "unicorn" ] && [ "${{.inputs.base_repo}}" = "ghcr.io/defenseunicorns/packages" ]; then
```
zachariahmiller, line 29:
```suggestion
          if [ "${{ .inputs.flavor }}" = "unicorn" ] && [ "${{ .inputs.base_repo }}" = "ghcr.io/defenseunicorns/packages" ]; then
```

tasks/utils.yaml
```diff
@@ -23,16 +22,11 @@ tasks:
...
             repo="${repo}/private"
           fi
           repo="${repo}/${{.inputs.team}}"
```
zachariahmiller, line 32:
```suggestion
          repo="${repo}/${{ .inputs.team }}"
```

templates/publish.yml
```diff
@@ -15,7 +15,7 @@ spec:
       default: ${IRON_BANK_ROBOT_PASSWORD}
     chainguard-identity:
       default: ${CHAINGUARD_IDENTITY}
-    target-repo:
```
zachariahmiller, line 18: Do we want to deprecate this instead of just hard deleting it since its technically a public interface and therefore a breaking change? It would need to be update in the .gitlab-ci.yml anywhere that is using it to continue working).

I'm similarly trying to think through if there is a case where we need to do this for the github pattern as well if someone was using TARGET_REPO directly. If so it would probably be the AI folks 🤔. We could put out feelers to determine if its actually being used anywhere. Otherwise, to be safe maybe we put a temporary conditional task with a deprecation warning and have it do:  `if TARGET_REPO != "" then set BASE_REPO=TARGET_REPO.`

zachariahmiller (changes requested):

templates/publish.yml
corang: Added logic to allow TARGET_REPO to still be set, gitlab logic was already redone to use the BASE_REPO var instead unless I missed something

zachariahmiller: Okay i think since the default .gitlab-ci.yml doesnt pass in that input explicitly by default we can avoid deprecating there and just swap to base_repo

ericwyles: Gitlab template is using TARGET_REPO here: https://github.com/defenseunicorns/uds-common/blob/4fa00d626863d7a9a118494ded28dd408e0ef97d/templates/publish.yml#L82

ericwyles: LOL, didn't realize I had made it to the part of the PR where you changed that hahaha looks ok

zachariahmiller (approved):

PR #405, opened 2025-02-06
feat: do not run commitlint on draft MRs

willswire:
## Description

commitlint shouldn't run on draft MRs in GitLab

## Checklist before merging

### DIFF
--- a/templates/commitlint.yml
+++ b/templates/commitlint.yml
@@ -8,7 +8,7 @@ spec:
 ---
 commitlint:
   rules:
-    - if: $CI_PIPELINE_SOURCE == "merge_request_event"
+    - if: $CI_PIPELINE_SOURCE == "merge_request_event" && $CI_MERGE_REQUEST_TITLE !~ /^Draft/
   script:
     - npm install --save-dev @commitlint/config-conventional@19.7.1
     - npm install --save-dev @commitlint/cli@19.7.1


willswire:
GitLab pipelines won't trigger when MR metadata changes (see https://gitlab.com/gitlab-org/gitlab/-/issues/25426 for example), however if a passing pipeline is required, users can just rerun the failed job once the title has changed.

I'll give it a whirl here on SEANET GitLab shortly by pointing to this ref.

zachariahmiller:
> GitLab pipelines won't trigger when MR metadata changes (see https://gitlab.com/gitlab-org/gitlab/-/issues/25426 for example), however if a passing pipeline is required, users can just rerun the failed job once the title has changed.
> 
> I'll give it a whirl here on SEANET GitLab shortly by pointing to this ref.

Sounds good and yeah that makes sense we're already tracking that issue it just wasnt in the forefront of my mind when i asked. Hopefully it gets resolved sooner rather than later. 

Since this is already a known gap requiring an upstream fix i think its kind of a moot point for the time being and as long as your testing doesn't bubble up some other unexpected behavior we should go ahead and make this change.

willswire:
No issues found in my tests.

zachariahmiller (approved):

zachariahmiller:
> No issues found in my tests.

I approved i'll wait until tomorrow in case anyone else wanted to weigh in before merging but should be good to go.

PR #407, opened 2025-02-06
feat: expose UDS CLI flag options for GitLab test template

willswire:
## Description

Allows for setting useful flags like `--insecure` or `--uds-cache` as inputs in the include section of GitLab CI/CD. 

## Checklist before merging


zachariahmiller (approved):
Makes sense to me and should preserve existing behavior unless you need to override the options.

PR #409, opened 2025-02-07
feat: add license config support

willswire:
## Description

Add the ability for repositories to use a `.license_config.yaml` file to set license information

## Checklist before merging


tasks/lint.yaml
```diff
@@ -159,56 +159,92 @@ tasks:
...
           darwin: bash
           linux: bash
         cmd: |
-          # check for addlicense bin
-          if [ -x "$HOME/go/bin/addlicense" ]; then
-              echo "addlicense installed in $HOME/go/bin"
+          CONFIG_FILE="license_config.yaml"
+
+          # If the config file exists, parse its values; otherwise, use default values.
+          if [ -f "$CONFIG_FILE" ]; then
```
mjnagel, line 170: Mosty a curiosity question - why make a new config file vs using task inputs/variables w/ defaults? I suspect that might need some adjustment to the callable workflow as well to really enable it (and I'm not sure what that looks like on the GitLab side) but might simplify the logic here and not require yaml parsing?

tasks/lint.yaml
```diff
@@ -159,56 +159,92 @@ tasks:
...
-      - description: Checking for SPDX headers
+      - description: Checking for SPDX headers using YAML config with defaults
         shell:
           darwin: bash
           linux: bash
         cmd: |
-          # check for addlicense bin
-          if [ -x "$HOME/go/bin/addlicense" ]; then
-              echo "addlicense installed in $HOME/go/bin"
+          CONFIG_FILE="license_config.yaml"
```
RothAndrew, line 167: did you want this to be `license_config.yaml`? or `.license_config.yaml`?

```suggestion
          CONFIG_FILE=".license_config.yaml"
```

tasks/lint.yaml
RothAndrew: I'm a fan of using the config file here. It's super easy to add and you can really easily determine whether the default AGPL stuff is being used or not in any particular repo by just checking for the presence of the .license_config.yaml file.

tasks/lint.yaml
```diff
@@ -159,56 +159,92 @@ tasks:
...
-      - description: Adding SPDX headers
+      - description: Adding SPDX headers using YAML config with defaults
         shell:
           darwin: bash
           linux: bash
         cmd: |
-          # check for addlicense bin
-          if [ -x "$HOME/go/bin/addlicense" ]; then
-              echo "addlicense installed in $HOME/go/bin"
+          CONFIG_FILE="license_config.yaml"
```
RothAndrew, line 212:
```suggestion
          CONFIG_FILE=".license_config.yaml"
```

tasks/lint.yaml
zachariahmiller: @mjnagel The optional config file was my suggestion. We want uds-common to be as simple and stable of an interface as possible and using a config file means we dont have to bubble up multiple different optional inputs through the tasks into the callable workflows into the parent workflows and similarly with gitlab components, which is more of a maintenance burden over time. It would also be easier to make additional changes or enhancements without needing to be as concerned about changes in each individual repo to support. 

I agree the yaml parsing in bash i less than ideal, but i think the longer term solution to that would be finding another way to define this task. This owuld have been a good use case to move the logic into uds-pk (formerly releaser) as we added additional capabilities, however the work on that tool has been stopped at least for now.

willswire: Concur w/ @zachariahmiller on this approach.

mjnagel: 🫡 sounds good and makes sense! And to be fair parsing a yaml file with yq is far less hacky than what we are already doing with gitignore 😆. Appreciate the context here.

tasks/lint.yaml
RothAndrew: looks like this one still needs to be fixed

zachariahmiller:
@willswire Could you please also add a small section in the docs that explains the way the licensing task and config file works?

willswire:
> @willswire Could you please also add a small section in the docs that explains the way the licensing task with the config file works?

Sure! Are you thinking in the root level README, or the tasks/README.md?

zachariahmiller:
> > @willswire Could you please also add a small section in the docs that explains the way the licensing task with the config file works?
> 
> Sure! Are you thinking in the root level README, or the tasks/README.md?

I was thinking about that before i asked and to be honest I'm not really sure. The root level README doesnt really get into any details like that and the tasks/README is actually just the output from `uds run --list`. It feels like the correct place is somewhere under the `docs/` directory, but none of the existing md files seem like a great fit and creating a new file just to describe this seems like overkill. @Racer159 @corang @mjnagel I'd welcome your opinions here on where this would fit best.

corang:
An addendum to the tasks readme makes sense to me

mjnagel:
It might get verbose but could we add to the task description (and likewise the task readme with the descriptions there). Someone can probably come up with a shorter form, but thinking along the lines of `uses .license_config.yaml or a default Defense Unicorns dual-license if not present`), possibly with a longer form explanation and example file in the task readme itself? If that seems lengthy (it probably is) then I'd also vote for just in the task readme.

tasks/lint.yaml
```diff
@@ -159,56 +159,92 @@ tasks:
...
-              exit 1
+            LICENSE="AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial"
+            COPYRIGHT="Defense Unicorns"
+          fi
+
+          # Exit if license or copyright are not provided.
+          if [ -z "$LICENSE" ] || [ "$LICENSE" = "null" ] || [ -z "$COPYRIGHT" ] || [ "$COPYRIGHT" = "null" ]; then
+            echo "License or copyright configuration not provided; exiting."
+            exit 0
+          fi
```
mjnagel, line 182: Should this `exit 0` if these are missing? Parsing the logic this would mean the user has a license-config file but is missing `license` and/or `copyright` keys. I would think that's likely a typo/misconfiguration and should error out?

If there's a use-case for explicitly having no license configuration I'm okay with the `null` or `none` fallbacks but should make sure that both license and copyright are set to that.

tasks/lint.yaml
```diff
@@ -159,56 +159,92 @@ tasks:
...
-            [ -z "$line" ] || [[ "$line" =~ ^# ]] && continue
-            # Add the line to ignore_flags array with quotes around it
-            ignore_flags+=("-ignore" "\"$line\"")
-          done < .gitignore
+          if [ -f "$CONFIG_FILE" ]; then
+            while IFS= read -r pattern; do
+              [ -z "$pattern" ] && continue
+              ignore_flags+=("-ignore" "\"$pattern\"")
+            done < <(uds zarf tools yq e '.ignore[]' "$CONFIG_FILE")
+          fi
```
mjnagel, line 197: I'm okay adding the new `ignore` option here in the config file but:
1. This ignore option would only be available if the end user is using a config file, which I think is _OK_ just means they have to explicitly add the 🦄 dual-license instead of using the default.
2. This removes the behavior of ignoring files from `.gitignore` automatically - forcing users to specify those same files in the new license config file. Unfortunately addlicense doesn't automatically ignore those which can be a pain for UDS Packages.

Could we keep the previous `.gitignore` behavior for backwards-compatibility? Otherwise I see this as a "breaking change" since users may be forced to create that config file and add all their gitignores to it.

mjnagel (approved):
LGTM - also appreciate the DRY-ing up of the check/fix steps.

zachariahmiller (approved):

PR #414, opened 2025-02-19
fix: pass down team input

willswire:
## Description

Pass down the `team` input to the publish task

## Checklist before merging


templates/publish.yml
```diff
@@ -80,7 +80,7 @@ publish:
 
         if [[ -n "$[[ inputs.base-repo ]]" ]]; then
           ARGS+=(--set BASE_REPO="$[[ inputs.base-repo ]]")
-        else
+        elif [[ -n "$[[ inputs.team ]]" ]]; then
```
zachariahmiller, line 83: @willswire Should this be an additional if statement instead of an `elif`? Technically both `base_repo` and `team` could both be appended to the args. Was there a reason you went to add it as `elif`?

willswire: whoops - yes!

zachariahmiller (approved):

PR #423, opened 2025-03-03
feat: create deploy template

willswire:
## Description

Provide a GitLab CI/CD Deployment component for doing deployments of dev packages.

## Checklist before merging

### DIFF
--- a/templates/deploy.yml
+++ b/templates/deploy.yml
@@ -1,0 +1,35 @@
+# Copyright 2024 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+spec:
+  inputs:
+    config:
+    environment:
+    runs-on:
+      default: gitlab-runner-4c-amd64
+    registry1-username:
+      default: ${IRON_BANK_ROBOT_USERNAME}
+    registry1-password:
+      default: ${IRON_BANK_ROBOT_PASSWORD}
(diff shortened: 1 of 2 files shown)


templates/deploy.yml
```diff
@@ -0,0 +1,35 @@
...
+                                              --set CHAINGUARD_TOKEN="$CHAINGUARD_TOKEN" \
+                                              --set GITLAB_REGISTRY_URL="$CI_REGISTRY" \
+                                              --set GITLAB_REGISTRY_USER="$CI_REGISTRY_USER" \
+                                              --set GITLAB_REGISTRY_TOKEN="$CI_REGISTRY_PASSWORD" \
+                                              --set GHCR_REGISTRY_USER="$GH_USER_READ_ONLY" \
+                                              --set GHCR_REGISTRY_TOKEN="$GH_PAT_READ_ONLY"
+    # Create package
+    - uds run create:package
+    # Deploy package
+    - uds run deploy:package --set ZARF_CONFIG="$[[ inputs.config ]]"
```
zachariahmiller, line 33: So I'm assuming the goal of this is to deploy the package to an external cluster. How are you managing the kube context? I suppose it could be in the environment already and "default", but should we provide a way to over ride that in a template like this?

zachariahmiller: I think im fine with that at least for an MVP of this. Can you update the readme with the example per your comment?

zachariahmiller (approved):

PR #426, opened 2025-03-04
feat: add Renovate support for Helm packages in the GitLab Package re…

willswire:
…gistry

## Description

...

## Checklist before merging


config/renovate.json5
```diff
@@ -273,6 +273,21 @@
             ],
             "datasourceTemplate": "docker"
         },
+        // Matches individual helm packages in a `zarf.yaml`'s `charts:` section that are stored in a GitLab package registry
```
zachariahmiller, line 276: to be honest i would have thought this would work out of the box with the existing patterns assuming you provided a [host_rules block](https://docs.renovatebot.com/modules/platform/gitlab/#setting-up-renovate-for-a-private-gitlab-container-registry) in your repo-specific renovate config that handled the instance endpoint and credentials. Did you try that first? Just asking...i can also see why it might not work, but i would have expected more in their documentation if it wasnt going to. 

Either way, can you please update this with a comment linking to a regex101 page where you have validated the pattern? That is a requirement we have whenever adding rules to the common config. For example https://github.com/defenseunicorns/uds-common/blob/dd794913f72452b641215b36744d22026334eca3/config/renovate.json5#L109

zachariahmiller (changes requested):

config/renovate.json5
zachariahmiller: Also @Racer159 and/or @mjnagel if you could please glance this over as well I'd appreciate it.

willswire: Turns out we just needed the following in our project's `renovate.json` `packageRules`:

```json
    {
      "matchDatasources": ["gitlab-tags", "gitlab-packages"],
      "registryUrls": ["https://gitlab.<our_gitlab_instance>.mil"]
    }
```

Because these datasources won't inherit the `RENOVATE_ENDPOINT`

PR #445, opened 2025-03-25
chore: switch all references from uds-releaser to uds-pk and use non deprecated cli commands

corang:
## Description

switch all references from uds-releaser to uds-pk and use non deprecated cli commands

## Checklist before merging

### DIFF
--- a/.github/workflows/callable-publish.yaml
+++ b/.github/workflows/callable-publish.yaml
@@ -28,7 +28,12 @@ on:
         default: 40
       uds-pk:
         type: boolean
+        default: true
+        description: "Whether to use the uds-pk for publishing"
+      uds-releaser:
+        type: boolean
         default: false
+        description: "DEPRECATED: Whether to use the UDS Releaser for publishing"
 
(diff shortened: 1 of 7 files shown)


.github/workflows/callable-publish.yaml
```diff
@@ -26,7 +26,7 @@ on:
       timeout:
         type: number
         default: 40
-      uds-releaser:
```
zachariahmiller, line 29: i think we should leave the input name as its slightly more descriptive and also changing it would be a breaking change

.github/workflows/callable-publish.yaml
```diff
@@ -26,9 +26,14 @@ on:
...
         type: number
         default: 40
+      uds-pk:
+        type: boolean
+        default: true
+        description: Whether to use the uds-pk for publishing
       uds-releaser:
         type: boolean
         default: false
+        description: "DEPRECATED: Whether to use the UDS Releaser for publishing"
```
mjnagel, line 36: Should this be a `deprecatedMessage`?

zachariahmiller (approved):

PR #446, opened 2025-03-25
feat: add workflow linting via step security api

corang:
## Description

Use the step security api ([website that can be used manually](https://app.stepsecurity.io/secure-workflow)) to check for and pin actions to shas

## Checklist before merging

### DIFF
--- a/tasks/lint.yaml
+++ b/tasks/lint.yaml
@@ -38,6 +38,7 @@ tasks:
       - task: shell
       - task: license
       - task: tasks
+      - task: workflows
 
   - name: yaml
     description: Run YAML linting checks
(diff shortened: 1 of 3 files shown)


tasks/lint.yaml
```diff
@@ -224,6 +224,47 @@ tasks:
           echo "Running: $cmd"
           eval "$cmd"
 
+  - name: workflows
+    description: Lint GitHub workflows
+    actions:
+      - shell:
+          linux: bash
+          darwin: bash
+        cmd: |
```
zachariahmiller, line 233: Using this approach i think this looks good, but i think we need to decide if we are okay with relying on the step security api or implementing similar logic to do the checks ourselves.

daveworth: I am *not* ok with this direction.

If we  call out to external APIs - we're setting ourselves up for flakey runs when the service is unavailable or rate limits us or makes any other breaking change.  I'd rather find ways to address this internally with `pinme` or something better.  Renovate is doing pretty well at helping us post-facto and a UDS Task or GitHub action with `pinme` (or something else) would be much better.

zachariahmiller:
Closing as we will go another direction if we look to implement (see comment from @daveworth)

PR #449, opened 2025-03-26
feat: adds a callable workflow for republishing a package

Michael-Kruggel:
## Description

Adds a workflow that will allow a user to republish a package. It will checkout the latest tag of a given flavor, create a zarf package from it, test a bundle using it, and redeploy over the latest tag of that flavor.

## Checklist before merging


corang (changes requested):

mjnagel (dismissed):
Could context be added to the description of this PR on why this toggle is needed / how it would be used? Generally I think we'd try to "roll forward" wherever possible for bugfixes (cut a new patch release) so I'm guessing this is for a different purpose?

I would need to review uds-pk code to double check but I assume this would also move the git tag/release to the latest commit as well?

zachariahmiller:
> Could context be added to the description of this PR on why this toggle is needed / how it would be used? Generally I think we'd try to "roll forward" wherever possible for bugfixes (cut a new patch release) so I'm guessing this is for a different purpose?
> 
> I would need to review uds-pk code to double check but I assume this would also move the git tag/release to the latest commit as well?

The use case for this is actually to be able to make the release mutable because in the case of ClamAV for example, the image with the virus definitions will be built and published nightly and as such, disconnected from version update, they want to be able to update the package to include the image with the latest definitions. Previously data injections were used, but that became insufficient for the disconnected environment use-case.

mjnagel:
> The use case for this is actually to be able to make the package mutable because in the case of ClamAV for example, the image with the virus definitions will be built and published nightly and as such, disconnected from version update, they want to be able to update the package to include the image with the latest definitions. Previously data injections were used, but that became insufficient for the disconnected environment use-case.

Is it true that in these case the image tag would not change (i.e. no zarf.yaml change needed), it's just the underlying sha that you want a newer image pulled? In other words - for this situation, is the intent ONLY to update the OCI published package, not the git tag/release?

If so I might lean towards this being a separate task/workflow where uds-pk isn't even involved, since this would exclusively be for zarf publish. That way there's no risk of this being used intentionally or accidentally to shift the underlying "code" for a given release.

zachariahmiller:
> > The use case for this is actually to be able to make the package mutable because in the case of ClamAV for example, the image with the virus definitions will be built and published nightly and as such, disconnected from version update, they want to be able to update the package to include the image with the latest definitions. Previously data injections were used, but that became insufficient for the disconnected environment use-case.
> 
> Is it true that in these cases the image tags would not change (i.e. no zarf.yaml change needed), it's just the underlying sha that you want a newer image pulled? In other words - for this situation, is the intent ONLY to update the OCI published package, not the git tag/release?
> 
> If so I might lean towards this being a separate task/workflow where uds-pk isn't even involved, since this would exclusively be for zarf publish. That way there's no risk of this being used intentionally or accidentally to shift the underlying "code" for a given release. In that scenario we'd want to checkout the existing git tag, rebuild/test, then publish over the same OCI tag.

That probably makes sense in which case it I'd debate whether this even needs to be in uds-common at all as this really is a more or less one off pattern we don't really want to propagate specific to the ClamAV package although I suppose there could be other similar use cases...

.github/workflows/republish-package.yaml
```diff
@@ -0,0 +1,42 @@
...
+  pull_request:
+    # milestoned is added here as a workaround for release-please not triggering PR workflows (PRs should be added to a milestone to trigger the workflow).
+    types: [milestoned, opened, reopened, synchronize]
+    # Will only rerun if files affecting this workflow are edited
+    paths:
+      - .github/workflows/callable-republish-package.yaml
+      - .github/workflows/republish-package.yaml
+      - tasks/actions.yaml
+      - tasks/create.yaml
+      - tasks/publish.yaml
```
mjnagel, line 17: I understand desired to test this on PR, but seems a bit sketchy to have permissions to publish/edit packages on PR (vs having it able to be called/on a cron)? But maybe not a big deal since this repo doesn't have "real" packages. Curious on @zachariahmiller's thoughts...

mjnagel (commented):
Looks good for the most part!

.github/workflows/republish-package.yaml
zachariahmiller: yeah this repo doesnt have real packages so i _think_ it should be fine. I actually think us not testing at least some level of the publish in this repo on PR is troubling because even last week for example it was objectively broken and we didnt know until @Michael-Kruggel ran into issues on this PR because there wasnt also an nginx update to trigger the release workflow when the breaking change was made. I for one am sick of having to deal with the fallout of deciding we wont test the release workflows. Had we not caught it early plenty of published packages would have silently become broken. 

I'm more than open to other suggestions though.

.github/workflows/callable-republish-package.yaml
```diff
@@ -0,0 +1,97 @@
...
+            uds run actions:setup-environment \
+            --set REGISTRY1_USERNAME="${{ secrets.IRON_BANK_ROBOT_USERNAME }}" \
+            --set REGISTRY1_PASSWORD="${{ secrets.IRON_BANK_ROBOT_PASSWORD }}" \
+            --set GH_TOKEN="${{ secrets.GITHUB_TOKEN }}" \
+            --set CHAINGUARD_IDENTITY="${{ secrets.CHAINGUARD_IDENTITY }}"
+        shell: bash
+
+      - name: Republish the package
+        run: |
+          uds run republish-package --set USE_CHECKPOINT=false --set FLAVOR=${{ inputs.flavor }} --no-progress ${{ inputs.options }}
```
corang, line 67: Why disable checkpoint?

tasks/create.yaml
```diff
@@ -31,6 +31,78 @@ tasks:
...
+              echo "No tag found for flavor '${{ .inputs.flavor }}'. Skipping."
+              exit 1
+          fi
+
+          uds-pk release update-yaml ${{ .inputs.flavor }}
+
+          if [ "${FLAVOR}" != "registry1" ] || [ "${{ .inputs.architecture }}" != "arm64" ]; then
+            ./uds zarf package create ${{ .inputs.path }} --confirm --no-progress --architecture="${{ .inputs.architecture }}" --flavor "${FLAVOR}" ${{ .inputs.options }}
+          else
+            echo "Registry1 packages cannot be made for 'arm64'"
```
corang, line 92: Should this have an `exit 1`?

corang (changes requested):

.github/workflows/republish-package.yaml
corang: Yeah for the sake of testing I think it being on PR with specific files makes sense, in a real repo it will likely be workflow_call and cron based

.github/workflows/callable-republish-package.yaml
corang: Talked to wayne, this is so we have a "full" test before publish in case checkpoint is being weird

corang (approved):

mjnagel (approved):

zachariahmiller (approved):

PR #462, opened 2025-04-09
feat: keycloak helper tasks

slaskawi:
## Description

This Pull Request introduces a few helper tasks for managing Keycloak

## Related Issue

Relates to: https://github.com/defenseunicorns/uds-identity-config/issues/373

## Checklist before merging

### DIFF
--- a/tasks/setup.yaml
+++ b/tasks/setup.yaml
@@ -57,36 +57,100 @@ tasks:
     description: Print the default keycloak 'admin' password to standard out (if INSECURE_ADMIN_PASSWORD_GENERATION was used on uds-core)
     actions:
       - description: Print the default keycloak admin password to standard out (if available)
-        cmd: ./uds zarf tools kubectl get secret -n keycloak keycloak-admin-password -o jsonpath='{.data.password}' | base64 -d
+        cmd: |
+          USERNAME=$(uds zarf tools kubectl get secret keycloak-admin-password -n keycloak -o jsonpath='{.data.username}' | base64 --decode)
+          PASSWORD=$(uds zarf tools kubectl get secret keycloak-admin-password -n keycloak -o jsonpath='{.data.password}' | base64 --decode)
 
-  - name: create-doug-user
-    description: Creates a user named 'doug' in the uds realm of keycloak (using the default admin account)
+          printf "\033[31m!!! Please ensure you're not running this in CI !!!\033[0m"
+          printf "Keycloak Admin Username: \033[31m %s \033[0m" "$USERNAME"
+          printf "Keycloak Admin Password: \033[31m %s \033[0m" "$PASSWORD"
+
+  - name: keycloak-admin-user
+    description: Sets up the Keycloak admin user for dev/testing if not already created
+    actions:
+      - description: Create Keycloak Admin User
+        cmd: |
+          # Check if the secret exists
+          if ./zarf tools kubectl get secret keycloak-admin-password -n keycloak > /dev/null 2>&1; then
+            echo "Admin user exists, skipping..."
+          else
+            # Start port-forward with zarf
+            ./zarf tools kubectl port-forward -n keycloak svc/keycloak-http 8080:8080 &
+            PF_PID=$!
+
+            # Wait a bit to ensure port-forward is ready
+            sleep 2
+
+            # Create admin user with curl
+            PASSWORD=$(openssl rand -base64 12)
+            STATE_COOKIE=$(curl -s --output /dev/null --cookie-jar - http://localhost:8080/ | grep "WELCOME_STATE_CHECKER" | awk '{print $7}')
+            curl -s --show-error http://localhost:8080/ \
+              -H "Cookie: WELCOME_STATE_CHECKER=${STATE_COOKIE}" \
+              -H "Content-Type: application/x-www-form-urlencoded" \
+              --data-urlencode "username=admin" \
+              --data-urlencode "password=${PASSWORD}" \
+              --data-urlencode "passwordConfirmation=${PASSWORD}" \
+              --data-urlencode "stateChecker=${STATE_COOKIE}"
+
+            # Kill the port-forward
+            kill $PF_PID
+
+            ./zarf tools kubectl create secret generic keycloak-admin-password \
+              --from-literal=username=admin \
+              --from-literal=password="${PASSWORD}" \
+              -n keycloak
+          fi
+
+  - name: keycloak-user
+    description: Creates a Keycloak user in the UDS Realm
     inputs:
-      keycloak_group:
+      group:
         description: Group to add user to
-        default: $KEYCLOAK_GROUP
+        default: $KEYCLOAK_USER_GROUP
+        required: false
+      username:
+        description: The username of the user to create
+        default: doug
+        required: false
+      password:
+        description: The password of the user to create
+        default: unicorn123!@#UN
+        required: false
+      first_name:
+        description: First name of the user to create
+        default: Doug
+        required: false
+      last_name:
+        description: First name of the user to create
+        default: Doug
         required: false
     actions:
-      - description: Creating the 'doug' user in the 'uds' realm
+      - task: keycloak-admin-user
+      - description: Creating the ${{ .inputs.username }} user in Keycloak
         cmd: |
-          KEYCLOAK_GROUP="${{ .inputs.keycloak_group }}"
+          KEYCLOAK_USER_GROUP="${{ .inputs.group }}"
+          KEYCLOAK_USER_NAME="${{ .inputs.username }}"
+          KEYCLOAK_USER_FIRST_NAME="${{ .inputs.first_name }}"
+          KEYCLOAK_USER_LAST_NAME="${{ .inputs.last_name }}"
+          KEYCLOAK_USER_PASSWORD="${{ .inputs.password }}"
+
           KEYCLOAK_ADMIN_PASSWORD=$(./uds zarf tools kubectl get secret -n keycloak keycloak-admin-password -o jsonpath='{.data.password}' | base64 -d)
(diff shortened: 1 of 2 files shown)


tasks/setup.yaml
```diff
@@ -57,36 +57,100 @@ tasks:
...
+          printf "Keycloak Admin Username: \033[31m %s \033[0m\n" "$USERNAME"
+          printf "Keycloak Admin Password: \033[31m %s \033[0m\n" "$PASSWORD"
+
+  - name: keycloak-admin-user
+    description: Sets up the Keycloak admin user for dev/testing if not already created
+    actions:
+      - description: Create Keycloak Admin User
+        cmd: |
+          # Check if the secret exists
+          if ./zarf tools kubectl get secret keycloak-admin-password -n keycloak > /dev/null 2>&1; then
```
zachariahmiller, line 74:
```suggestion
          if ./uds zarf tools kubectl get secret keycloak-admin-password -n keycloak > /dev/null 2>&1; then
```

Inside a task it has to be `./uds zarf` you can only use `./zarf` in a zarf.yaml

tasks/setup.yaml
```diff
@@ -57,36 +57,100 @@ tasks:
...
+            echo "Admin user exists, skipping..."
+          else
+            # Start port-forward with zarf
+            ./zarf tools kubectl port-forward -n keycloak svc/keycloak-http 8080:8080 &
```
zachariahmiller, line 78:
```suggestion
            ./uds zarf tools kubectl port-forward -n keycloak svc/keycloak-http 8080:8080 &
```

tasks/setup.yaml
```diff
@@ -57,36 +57,100 @@ tasks:
...
+              -H "Content-Type: application/x-www-form-urlencoded" \
+              --data-urlencode "username=admin" \
+              --data-urlencode "password=${PASSWORD}" \
+              --data-urlencode "passwordConfirmation=${PASSWORD}" \
+              --data-urlencode "stateChecker=${STATE_COOKIE}"
+
+            # Kill the port-forward
+            kill $PF_PID
+
+            ./zarf tools kubectl create secret generic keycloak-admin-password \
```
zachariahmiller, line 98:
```suggestion
            ./uds zarf tools kubectl create secret generic keycloak-admin-password \
```

zachariahmiller (changes requested):
In general i think this looks fine. Requested a couple of changes and asked a couple of questions. Inclined to approve this if the answers are "yes" and if you can make the very minor changes i suggested.

tasks/setup.yaml
```diff
@@ -57,36 +57,100 @@ tasks:
...
+            curl -s --show-error http://localhost:8080/ \
+              -H "Cookie: WELCOME_STATE_CHECKER=${STATE_COOKIE}" \
+              -H "Content-Type: application/x-www-form-urlencoded" \
+              --data-urlencode "username=admin" \
+              --data-urlencode "password=${PASSWORD}" \
+              --data-urlencode "passwordConfirmation=${PASSWORD}" \
+              --data-urlencode "stateChecker=${STATE_COOKIE}"
+
+            # Kill the port-forward
+            kill $PF_PID
```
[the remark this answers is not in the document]
mjnagel: We currently use this task in [uds-core CI](https://github.com/defenseunicorns/uds-core/blob/main/tasks/utils.yaml#L33-L66) for any non-bundle deployments (where the insecure admin is not setup). We've had it in our repo for ~6 months and it's gone through hundreds++ of pipelines + local deploys I'd imagine, without any failures to my knowledge.

tasks/setup.yaml
mjnagel: `./zarf` actually should still work in tasks (it does for our uds-core version of this, although `./uds zarf` is probably better for consistency sake.

zachariahmiller: Okay that's good to know i think i tracked down the previous conversation that was making me think it wouldn't. Most of the confusion around this in the past was actually related to folks trying to use `./uds zarf` within a zarf.yaml.

In any case, for consistency i would still ask they get changed to `./uds zarf`

tasks/setup.yaml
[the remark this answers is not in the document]
slaskawi: Yes, just to echo what @mjnagel said, this task is almost exact copy-paste from UDS Core. So far we haven't seen any inconsistencies but if anything is there, we'll be using this in UDS Identity Config as well and we'll have more chances to catch it.

tasks/setup.yaml
```diff
@@ -57,36 +57,100 @@ tasks:
...
+          # Check if the secret exists
+          if ./zarf tools kubectl get secret keycloak-admin-password -n keycloak > /dev/null 2>&1; then
+            echo "Admin user exists, skipping..."
+          else
+            # Start port-forward with zarf
+            ./zarf tools kubectl port-forward -n keycloak svc/keycloak-http 8080:8080 &
+            PF_PID=$!
+
+            # Wait a bit to ensure port-forward is ready
+            sleep 2
```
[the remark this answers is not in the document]
slaskawi: We used to use 5 here. I tested this locally a few times and lowering was fine. But maybe you're right, I'll revert it to 5 as an initial commit. We can always lower it down.

zachariahmiller (approved):

slaskawi:
Thanks for the review @zachariahmiller !

I'll probably test it a bit more in UDS Core before the merge. I want to be 100% sure I won't break anything and we're safe to move Keycloak things here.

tasks/setup.yaml
```diff
@@ -57,36 +57,100 @@ tasks:
     description: Print the default keycloak 'admin' password to standard out (if INSECURE_ADMIN_PASSWORD_GENERATION was used on uds-core)
     actions:
       - description: Print the default keycloak admin password to standard out (if available)
-        cmd: ./uds zarf tools kubectl get secret -n keycloak keycloak-admin-password -o jsonpath='{.data.password}' | base64 -d
+        cmd: |
+          USERNAME=$(./uds zarf tools kubectl get secret keycloak-admin-password -n keycloak -o jsonpath='{.data.username}' | base64 --decode)
+          PASSWORD=$(./uds zarf tools kubectl get secret keycloak-admin-password -n keycloak -o jsonpath='{.data.password}' | base64 --decode)
 
-  - name: create-doug-user
```
UnicornChance, line 62: If we want to rename this task, we need to make sure with is a noted as a breaking change. [Quick search shows that 32 different files](https://github.com/search?q=org%3Adefenseunicorns+%22create-doug-user%22+language%3AYAML&type=code&l=YAML) in the company github reference this task, so it will impact a lot of people.

UnicornChance: oops, i didn't scroll far enough to see that the create-doug-user was maintained at the bottom of the file...

UnicornChance (dismissed):
LGTM. I don't feel strongly about this, but does feel weird to introduce a new task rather than just update the existing `create-doug-user` functionality.

tasks/setup.yaml
```diff
@@ -57,36 +57,100 @@ tasks:
...
+        description: The password of the user to create
+        default: unicorn123!@#UN
+        required: false
+      first_name:
+        description: First name of the user to create
+        default: Doug
+        required: false
+      last_name:
+        description: First name of the user to create
+        default: Doug
```
mjnagel, line 125: extremely nit: `doug` user's last name was Unicorn previously. I highly doubt anything is doing validations on this, but may be good to keep the default the same just in case.

tasks/setup.yaml
```diff
@@ -96,23 +160,35 @@ tasks:
...
           --header "Content-Type: application/json" \
-          --header "Authorization: Bearer ${KEYCLOAK_ADMIN_TOKEN}" \
+          --header "Authorization: Bearer ${KEYCLOAK_ADMIN_ACCESS_TOKEN}" \
           --data "{
                   \"id\": \"${CONDITIONAL_OTP_ID}\",
                   \"requirement\": \"DISABLED\"
               }"
+
+  - name: create-doug-user
+    description: DEPRECATED! Use keycloak-user instead
```
mjnagel, line 185: Should we actually deprecate this? If we're not urgently removing it I guess that's fine since it's just words, but @zachariahmiller n co might have more work to switch off of this task everywhere so that they aren't using something documented as deprecated.

slaskawi:
> LGTM. I don't feel strongly about this, but does feel weird to introduce a new task rather than just update the existing create-doug-user functionality.

@UnicornChance The primary reason I deprecated it is that the task name falls out of the naming convention. All Keycloak related tasks start with `keycloak` and this one didn't and it was harder to find in my opinion.

I left it there just in case someone uses it. I'll probably remove it in a bit.

tasks/setup.yaml
slaskawi: aaah you're eagle-eyed @mjnagel ! Fixed!

tasks/setup.yaml
slaskawi: Unfortunately this is quite heavily used across the other packages. So we can't just remove it. But I wanted to at least prevent it from creeping out and let people know about the new, more flexible task. I'm open if there are better ways to do it.

zachariahmiller: I was thinking we could just leave it as "deprecated" for an indeterminate amount of time and in that case I am fine with the change, but I suppose we could remove the note that its deprecated and keep both or use wording like "consider using keycloak-user instead". 

To be honest almost all the apps we test just need a single user for automated ui testing and that is why this consistent "doug" user exists how it is. 

I'm not keen on having to go update everywhere its used and we don't really have any hard and fast naming conventions here, but since its been updated to just be a thin wrapper on the new generic task, having them both there isn't really any additional overhead. 🤷

slaskawi: @zachariahmiller Updated the comment description and turned on automerge.

mjnagel (approved):

PR #468, opened 2025-04-15
fix: create-doug-user calls wrong task

corang:
## Description

...

## Checklist before merging


zachariahmiller (approved):
thanks for fixing this.

PR #469, opened 2025-04-16
feat: allow publish path to be set by the user

JeffResc:
## Description
Adds a `publish-path` input to [`utils:determine-repo`](https://github.com/defenseunicorns/uds-common/blob/e84f5624ec53b7e126166ae4d9f1144ba8afeac6/tasks/utils.yaml#L10-L54) that will ignore the existing path building logic and use the user-set value as a path for publishing.

A limitation of the current approach is that [GitLab container registry allows a maximum depth of two in the publish path](https://docs.gitlab.com/user/packages/container_registry/#naming-convention-for-your-container-images). The current implementation works great on upstream and registry1 variants as they are published to `/uds/mypackage`, but publishes of unicorn flavors are published to `/uds/private/mypackage`, which does not comply with GitLab's naming convention resulting in a 401 when publishing.

The `publish-path` input allows the user to set their own publish path to overcome this limitation.

## Checklist before merging

### DIFF
--- a/tasks/publish.yaml
+++ b/tasks/publish.yaml
@@ -12,6 +12,7 @@ variables:
   - name: ENABLE_UDS_PK
     default: "false"
   - name: BASE_REPO
+  - name: PUBLISH_PATH
 
 tasks:
   - name: package
@@ -72,6 +73,7 @@ tasks:
         with:
           team: ${{ .variables.TEAM }}
           base_repo: ${{ .variables.BASE_REPO }}
+          publish_path: ${{ .variables.PUBLISH_PATH }}
       - description: Get the current Zarf package name
         cmd: cat ${{ .inputs.path }}/zarf.yaml | ./uds zarf tools yq .metadata.name
         setVariables:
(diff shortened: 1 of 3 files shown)


JeffResc:
How should [this lint test](https://github.com/defenseunicorns/uds-common/actions/runs/14495357260/job/40661578165?pr=469#step:6:40) be addressed? It's resolving `${{ .inputs.publish_path }}` as `maru-expression`, then saying that `-n "maru-expression"` will always be `true`. That's correct, but `${{ .inputs.publish_path }}` could also have a value of `""`.

zachariahmiller:
> How should [this lint test](https://github.com/defenseunicorns/uds-common/actions/runs/14495357260/job/40661578165?pr=469#step:6:40) be addressed? It's resolving `${{ .inputs.publish_path }}` as `maru-expression`, then saying that `-n "maru-expression"` will always be `true`. That's correct, but `${{ .inputs.publish_path }}` could also have a value of `""`.

like this https://github.com/defenseunicorns/uds-common/blob/42196a596353dc9f8acb21a6a86b5d4a4f7ee76f/tasks/create.yaml#L81

PR #491, opened 2025-05-20
fix: make verify badge in ci never cause a failure

corang:
## Description

...

## Checklist before merging


zachariahmiller (approved):
I think this is fine until such a time where we might want to actually enforce compliance of, rather than just check the requirements.

PR #512, opened 2025-06-23
chore: install-deps, uds-pk update-yaml, zarf yq test template update

anthonywendt:
## Description
Making some need test template updates based on usage in customer environment.

- Updated to use zarf tools yq
- Added actions:install-deps
- Added uds-pk update-yaml

### DIFF
--- a/templates/test.yml
+++ b/templates/test.yml
@@ -45,7 +45,7 @@ test:
         git checkout "$TAG"
 
         # Extract upgrade flavors and set UPGRADE_FLAVORS
-        UPGRADE_FLAVORS=$(cat $[[ inputs.check-flavor-zarf-yaml ]] | yq '.components[] | select(.only | has("flavor")) | .only.flavor' | paste -s -d, - || echo "")
+        UPGRADE_FLAVORS=$(cat $[[ inputs.check-flavor-zarf-yaml ]] | uds zarf tools yq '.components[] | select(.only | has("flavor")) | .only.flavor' | paste -s -d, - || echo "")
 
         # change back to the commit branch for the rest of the job
         git checkout -
@@ -71,6 +71,10 @@ test:
                                               --set GITLAB_REGISTRY_TOKEN="$[[ inputs.ci-registry-password ]]" \
                                               --set GHCR_REGISTRY_USER="$GH_USER_READ_ONLY" \
                                               --set GHCR_REGISTRY_TOKEN="$GH_PAT_READ_ONLY"
+
+    - sudo uds run actions:install-deps
+    # Modify zarf package and uds bundle to have the correct version
+    - uds-pk update-yaml $[[ inputs.flavor ]]
     - uds run actions:test-deploy --set FLAVOR="$[[ inputs.flavor ]]" --set TYPE="$[[ inputs.type ]]" --set OPTIONS="$[[ inputs.options ]]"
     - uds run actions:verify-badge
 


templates/test.yml
```diff
@@ -71,6 +71,10 @@ test:
                                               --set GITLAB_REGISTRY_TOKEN="$[[ inputs.ci-registry-password ]]" \
                                               --set GHCR_REGISTRY_USER="$GH_USER_READ_ONLY" \
                                               --set GHCR_REGISTRY_TOKEN="$GH_PAT_READ_ONLY"
+
+    - sudo uds run actions:install-deps
+    # Modify zarf package and uds bundle to have the correct version
+    - uds-pk update-yaml $[[ inputs.flavor ]]
```
zachariahmiller, line 77: this should be `uds-pk release update-yaml`, but why do you actually need this? It probably doesnt matter either way just wondering.

templates/test.yml
```diff
@@ -71,6 +71,10 @@ test:
                                               --set GITLAB_REGISTRY_TOKEN="$[[ inputs.ci-registry-password ]]" \
                                               --set GHCR_REGISTRY_USER="$GH_USER_READ_ONLY" \
                                               --set GHCR_REGISTRY_TOKEN="$GH_PAT_READ_ONLY"
+
+    - sudo uds run actions:install-deps
```
zachariahmiller, line 75: This should realistically not have sudo. What is the makeup of the runner (executor, etc) that is running these job templates?

zachariahmiller (changes requested):

templates/test.yml
zachariahmiller: This also shouldnt necessarily need to run. For example all fleeting based runners would already have the dependencies directly on the host. Docker based runner should probably have a a purpose built container where these are already installed. At best this should be an optional step. Open to hearing more on your use case.

templates/test.yml
anthonywendt: The publish template was doing this step and when we brought in the test job usting the test tempate the zarf and bundle versions were not right and would fail. So adding this to modify those to the correct version fixed that in the test template and made it match the same workflow that the publish task is using.

templates/test.yml
zachariahmiller: I would be open to adding a container build to the same project where we build the fleeting AMIs. This process was built around the assumption that folks would be using the VM based fleeting runners. If docker is being used imo a container should be getting built with the dependencies separately and consumed in the pipeline, they shouldnt be getting installed at ci time, the amount of time wasted by doing that is a lot over time and you cant always ensure you have the same versions of things.

templates/test.yml
anthonywendt: We could just set the versions to dev in the bundle and zarf package, but was tryin got just be consistent across templates

templates/test.yml
zachariahmiller: Well we can assume that folks will only use this repo with the fleeting VM runners or the docker-autoscaler fleeting runners that we provide a package for or at least that those are the only types we support. In both cases the deps should already exist either on the VM or the image. We cant and wont support every way someone might deploy their own gitlab runners. For example never supporting a powershell based shell runner, custom executor runner, docker-machine runner, virtualbox runner, etc.  

Similarly, this running with a k8s executor would require using our package (or the same config) that uses podman to allow non-root deployment of k3d. 

I'm strongly of the opinion that generally speaking, folks should build or consume minimal, purpose-built images for the CI jobs they are running and often this requires managing them themselves as they are the only ones who know all the tools and specific versions they need for their use-case. A "kitchen sink" sort of mega build container with everything imaginable on it is also an anti-pattern and while i do think us producing the image we are talking about is approaching that, it seems fine to me to do that in order to eliminate the initial chicken/egg situation with getting access to our tooling. 

All that to say:

1. We will and do assume that someone is deploying one of our supported gitlab runner configurations (docker-autoscaler fleeting, instance (vm) fleeting or k8s configured to run rootless podman)
2. We should build this "uds" image to at least bridge the gap for running with all of our tools, but i wouldn't go as far as to say it would be a best practice to then use that image as is for everything.

Can you please create an issue [here](https://github.com/defenseunicorns/uds-image-builder) for creating that image? I'll bring it up to the team to look at getting it implemented

anthonywendt: I appreciate the time and thought. This makes sense to me. I can create that issue

anthonywendt: So I updated this to add --preserve-env to uds run actions:install-deps. This allows it to work with mirrored versions of uds-common. 

They are also using docker runners that will require sudo for this to work. We intend on creating our own image that will have all the things we need installed but feel weird about just removing this call to actions:install-deps because others could potentially be using it and relying on it? It has been in this publish.yml template since before I messed with it.

templates/test.yml
zachariahmiller: The pattern we have established for using the releaser with packages was that the releaser.yaml is the source of truth for package versions and packages/local test bundle would just be "dev". I don't really see a good reason to change that. I'm open to being persuaded otherwise and may be missing something in my mental model since I dont have an active gitlab env running these anymore, but as it stands deviating from that without any other reason seems like making the process more complicated with little to no gain.

templates/test.yml
ericwyles: Looking at this more, install-deps is only installing k3d, lula, and uds-pk. If you set your version numbers to `dev` then the uds-pk isn't really needed. Do you actually still even need install-deps for your use case?

templates/test.yml
anthonywendt: Oh I was just trying to be consistent with what is already in these templates. I copied what was being done in the publish template. If that is the established pattern then I agree these templates need to all reflect that pattern. Just didn't know that was an established pattern. It seems these gitlab templates need some attention from the powers that be. We were just trying to use them. 

I'll update the customer package to use dev for version and can delete the uds-pk call from the templates all together.

templates/test.yml
anthonywendt: I appreciate this. We actually did this in our before script initially when we first started. What happened was we realized that the publish template here in uds-common worked fine and it was because it was calling install-deps. Pretty much the only reason I updated the test template was to make it consistent with what is in the publish template.

I see a theme between these comment threads being that these gitlab templates need some work in general to represent the expectations that I am hearing in these comments. I agree with all these comments, just noting that I was modeling what the other template(publish) in uds-common was doing.

templates/test.yml
anthonywendt: And by we were trying to just use them I mean a customer is trying to use them in their own software factory with their own runners.

anthonywendt:
Updated to cleanup the publish and test template based on all the conversation. Testing in customer environment with the necessary changes on their end to work with this cleanup.

zachariahmiller:
Will be superseded by this PR https://github.com/defenseunicorns/uds-common/pull/529

PR #523, opened 2025-07-18
feat: add verify job to simplify required CI checks

corang:
## Description

Reworks the test workflow to no longer need a shim workflow and have a specific job that is consistently named to be referenced for required checks

## Checklist before merging


zachariahmiller (commented):
@corang A few thoughts/ideas on here. 

1. As is this would need to be callable to be consumed in each repo.
2. Should probably have a timeout and probably some error handling around if there are network blips or other issues on the gh cli api calls
3. As is should probably be filtered to only check against the test jobs. We can add the lint and pr checks generically to required checks as it is. I dont think we want the scan workflow failing to block PRs. I dont think this really matters much either way how this is accomplished, but we wouldnt want to depend on the scan jobs completing/succeeding at least at this time. 

Alternatively, what if this was changed around slightly to be a callable workflow that is run as a final step on each repo test.yaml as an `if: always()` and `needs: validate`? Then it could have the matrix passed into is so it is only querying jobs generated by the matrix or if that is clunky/flaky then still do what I'm suggesting process-wise, but omit passing in the matrix and just check any job names like you are doing now, but at the end of the test workflow after the validate job/s have completed.

That would still accomplish our goals while also allowing the validation it is doing to be more precise in checking the completion of just the test jobs and it would also eliminate the need to do the polling throughout the entire duration of all of the ci jobs running. 

Thinking about this a little more, there may be cases where a repo has a subset of non-standard jobs that should be required (ie deploy into a cloud and test something), so maybe we need to figure out a way to handle that, potentially an input of jobs to omit and then it checks the status of all the rest as being completed.

.github/workflows/callable-test.yaml
```diff
@@ -37,18 +37,88 @@ on:
         # renovate: datasource=github-tags depName=defenseunicorns/uds-cli versioning=semver
         default: 0.27.8
         type: string
+      file-filter:
+        type: string
+        description: |
+          A string to filter files that should be tested.
```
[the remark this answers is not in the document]
corang: I think it's better now

zachariahmiller (approved):

.github/workflows/callable-test.yaml
```diff
@@ -37,18 +37,87 @@ on:
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          FILE_FILTER_INPUT: ${{ inputs.file-filter }}
+        run: |
+          # allow ** for recursive globs
+          shopt -s globstar
+
+          MODIFIED_FILES=$(gh pr view ${{ github.event.pull_request.number }} \
+            --repo ${{ github.repository }} --json files \
+            --jq '.files[].path')
```
mjnagel, line 89: This is probably fine since we run CI on each commit, but there might be some timing here where `gh pr view` gets a different set of files/diffs than the commit that actually got pushed (this is an API call rather than looking at the current commit diffs).

Alternative here might be something like `git diff --name-only ${{ github.event.pull_request.head.sha }}` (I think that's the right var), which would get the list of files changed from the current commit back to the head of the PR.

mjnagel: The timing I'm thinking of is actually two commits in quick succession. The first commit might get the diffs from the second commit, because the PR has new commits. Again probably irrelevant since we cancel CI when newer commits happen. 

For what its worth dorny/paths-filter uses `git` commands only I think ([ref](https://github.com/dorny/paths-filter/blob/master/src/git.ts)). I don't think it's using the gh CLI/API?

PR #524, opened 2025-07-18
feat: uds registry badging requirements

fingermustache:
Document what it means for a package to be ready for the UDS Registry. These requirements are basically what we used to call the "silver tier."

### What needs to happen:

- [x] Create a doc that lists all the UDS Registry Package Requirements (based on the old silver tier).
- [x]  Update the wording so that requirements that aren’t always possible (due to upstream or app-specific reasons) use "should" instead of "must" — especially stuff that used to be strict in the gold tier.
- [x]  Remove all mentions of bronze, silver, and gold — just refer to everything as UDS Registry Package Requirements from now on.
- [ ]  Add a note that the "Made for UDS" indicator should only be added to the package README once a package meets all the requirements.

This will make things clearer going forward and help us standardize how we vet packages for the registry.

## Checklist before merging


codyshoffner (approved):

fingermustache:
- [ ] Add a note that the "Made for UDS" indicator should only be added to the package README once a package meets all the requirements.
 
 This will be added to the README in templates

PR #656, opened 2026-04-23
chore: add package schema to vscode settings.json

jameslanska-du:
## Description

- Add UDS Package schema to template VSCode settings.json to highlight deviations from the schema and enable docs on hover.
- Standardize indentation (all lines but the first were previously indented an extra two spaces)

## Checklist before merging

### DIFF
--- a/config/.vscode/settings.json
+++ b/config/.vscode/settings.json
@@ -20,7 +20,8 @@
       "zarf.yaml"
     ],
     "https://raw.githubusercontent.com/defenseunicorns/uds-core/main/schemas/package-v1alpha1.schema.json": [
-      "package.yaml"
+      "package.yaml",
+      "uds-package.yaml"
     ]
   },
   "cSpell.words": [
--- a/config/.vscode/settings.json
+++ b/config/.vscode/settings.json
@@ -20,9 +20,14 @@
       "zarf.yaml"
     ],
     "https://raw.githubusercontent.com/defenseunicorns/uds-core/main/schemas/package-v1alpha1.schema.json": [
-      "package.yaml",
-      "uds-package.yaml"
-    ]
+      "**/package.yaml",
+      "**/uds-package.yaml"
+    ],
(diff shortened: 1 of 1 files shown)


config/.vscode/settings.json
```diff
@@ -1,43 +1,46 @@
...
+    "https://raw.githubusercontent.com/defenseunicorns/uds-cli/v0.30.3/tasks.schema.json": [
+      "tasks.yaml",
+      "tasks/**/*.yaml",
+      "src/**/validate.yaml"
+    ],
+    "https://raw.githubusercontent.com/defenseunicorns/zarf/v0.75.0/zarf.schema.json": [
+      "zarf.yaml"
+    ],
+    "https://raw.githubusercontent.com/defenseunicorns/uds-core/main/schemas/package-v1alpha1.schema.json": [
+      "package.yaml"
```
mjnagel: super nit: would probably add `uds-package.yaml` here. I think that's a common filename, maybe even more common than `package.yaml`.

zachariahmiller: Also will this actually work as you intend it to? First, the uds-package.yaml is relative to the chart/charts directory. Second, its in every case i have ever seen actually a helm template not a standard yaml file and i dont think that will work with the json schema and will just show a bunch of errors.

jameslanska-du: That is a great question!

Yes.  It will work as I intended for YAML files, and it is a no-op for Helm templated UDS Packages.  See https://github.com/defenseunicorns/uds-core/blob/main/.vscode/settings.json#L25 for reference.

For `.yaml` files with `{{ }}` syntax, VSCode sets the language-id to be "helm-template" which is not an activation event for the redhat.vscode-yaml VSCode extension.  Therefore, no checking is performed.  The helm-ls vscode extension theoretically could enable this functionality, but it appears to have only a couple thousand downloads and does not appear to be functional at the moment.

Secondly, the `**/` syntax is not necessary to find files nested in subdirectories.  However, I have added it for improved readability and consistency with UDS Core.

PR #679, opened 2026-05-20
chore: add Zarf values guidelines for UDS Packages

Racer159:
## Description

This adds Zarf values guidance and wiring for implementing this feature within UDS Packages going forward.

## Checklist before merging


docs/uds-packages/guidelines/values-guidelines.md
```diff
@@ -0,0 +1,238 @@
...
+          - sourcePath: .uds-nginx-config
+            targetPath: .
+      - name: nginx
+        version: 0.1.0
+        namespace: nginx
+        localPath: ../src/nginx
+        valuesFiles:
+          - ../values/common-values.yaml
+        values:
+          - sourcePath: .nginx.replicaCount
```
joelmccoy, line 43: do you imagine if we get the [negative pattern ](https://github.com/zarf-dev/zarf/issues/4878#issuecomment-4408679576) in zarf values you would recommend just doing a `.nginx. -> .` here and just put negative's on things like security context, image, etc?

docs/uds-packages/guidelines/values-guidelines.md
```diff
@@ -0,0 +1,238 @@
...
+
+The schema file should follow the [JSON Schema](https://json-schema.org/) specification and should be best effort to validate the common values that a consumer of the UDS Package would reasonably expect to have configured.  For deeply nested values the UDS Package should seek to keep the schema simple and avoid over-complicating it as this could cause unexpected breakages as the underlying Helm chart or Kubernetes API changes.
+
+The `values` mappings (`source` and `target`) to Helm charts should be added alongside the initial Helm chart definition (i.e. in the `common/zarf.yaml` file).  These should be kept as flat as possible, exposing the highest order node in the values tree that makes sense for the package.  Given most UDS Packages are single-component, the top level mapping should be the chart's name and should exclude the component name for simplicity (if a multi-component package needs it component name can be added as necessary).
```
joelmccoy: What if we have a value that can be shared across multiple charts/components? A very simple example is could be the `Domain` being shared across the app chart and the uds-config-chart.  Do we recommend we have two separate name-spaced values for this? or would we want to have this as a top level shared value:

namespaced:
```yaml
      - name: uds-nginx-config
        namespace: nginx
        version: 0.1.0
        localPath: ../chart
        values:
          - sourcePath: .uds-nginx-config.domain
            targetPath: .domain
      - name: nginx
        version: 0.1.0
        namespace: nginx
        localPath: ../src/nginx
        valuesFiles:
          - ../values/common-values.yaml
        values:
          - sourcePath: .nginx.domain
            targetPath: .domain
```

shared:
```yaml
      - name: uds-nginx-config
        namespace: nginx
        version: 0.1.0
        localPath: ../chart
        values:
          - sourcePath: .shared.domain
            targetPath: .domain
      - name: nginx
        version: 0.1.0
        namespace: nginx
        localPath: ../src/nginx
        valuesFiles:
          - ../values/common-values.yaml
        values:
          - sourcePath: .shared.domain
            targetPath: .domain
```

joelmccoy (dismissed):
Generally on board with the guidance here! Just two small questions

docs/uds-packages/guidelines/values-guidelines.md
```diff
@@ -0,0 +1,238 @@
+# UDS Package Values Guidelines
+
+When implementing UDS Packages, it is important to consider how [Zarf Values](https://docs.zarf.dev/ref/examples/values-templating/) will be exposed from the package and what the implications are for downstream consumers.  UDS Packages are expected to be deployed in a variety of environments, and the values exposed through to Helm charts should be flexible enough to accommodate different deployment scenarios while also being secure and easy to use.
+
+This means that for any given UDS Package most of the values from the underlying Helm chart should be exposed as Zarf Values with only values that would inherently break the UDS Package (i.e. `image` references, or `ingress` configurations) or compromise the security of the package (i.e. `podSecurityContext`) being restricted.
+
+## Implementation
+
+The top-level `values` key for a UDS Package should be defined in the parent `zarf.yaml` of the UDS Package repository and should reference a `zarf-default-values.yaml` file and a `zarf-default-values.schema.json` file located in the root of the repository to handle defaults and validation respectively.
```
zachariahmiller: Wouldn't it make sense for zarf to be able to generate this schema given the defined mapping and the underlying structures that are expected rather than having to handcraft/curate and maintain these schemas on a per package basis?

zachariahmiller: Its been a while since reviewing the zarf values proposal/implementation details so bear with me... Default values as in shipped defaults (analogous to bundle variables with defaults set in the bundle.yaml) or just an example, not included in the packaged artifact (analogous variable override definition in a uds-config.yaml with a bundle.yaml with variable overrides and no defaults)? 

If the latter this seems reasonable. If the former i would be concerned about who they are actually good defaults for since they are getting baked in. In that case i would want something more like named sets of defaults that are included and could be referenced/applied.

docs/uds-packages/requirements/uds-package-requirements.md
```diff
@@ -26,22 +26,7 @@ UDS Packages may make use of the [UDS Exemption custom resource](https://github.
...
-- [ ] **Should** clearly mark the Keycloak client id with the group and app name `uds-<group>-<application>` (i.e. `uds-swf-mattermost`) to provide consistency in the Keycloak UI.
-- [ ] **Should** limit the use of Zarf variable templates and prioritize configuring packages via Helm value overrides.
-  > This ensures that the package is configured the same way that the bundle would be and avoids any side effect issues of Zarf's `###` templating.
-- [ ] **May** template network policy keys to provide flexibility for delivery customers to configure.
-- [ ] **May** end any generated Keycloak client secrets with `-sso` to easily locate them when querying the cluster.
-- [ ] **May** template Keycloak fields to provide flexibility for delivery customers to configure.
-- [ ] **Should** be created from the [UDS Package Template](https://github.com/uds-packages/template)
-- [ ] **Must** be declaratively bundled in a [Zarf package](https://docs.zarf.dev/ref/create/)
+- [ ] **Must** be declaratively packaged in a [Zarf package](https://docs.zarf.dev/ref/create/) that is deployable via [`zarf package deploy`](https://docs.zarf.dev/ref/deploy/) (not another tool).
```
zachariahmiller: feels like a weird reframing trying to imply "not with bundles", but is still isnt expected to deploy unless its in a cluster with uds-core deployed so just zarf is insufficient realistically.

docs/uds-packages/requirements/uds-package-requirements.md
```diff
@@ -57,8 +42,23 @@ UDS Packages may make use of the [UDS Exemption custom resource](https://github.
   > If web requests need to be made they should be done through a `Job` or `./zarf tools kubectl exec` as appropriate.
 - [ ] **Must** not use local commands outside of `coreutils` or `./zarf` self references within `actions`.
 - [ ] **Must** include application [metadata for UDS Registry](https://github.com/defenseunicorns/uds-common/blob/main/docs/uds-packages/guidelines/metadata-guidelines.md) publishing
+- [ ] **Should** expose all configuration (`uds.dev` CRs, additional `Secrets`/`ConfigMap`s, etc) through a Helm chart (ideally in a `chart` or `charts` directory).
+  > This allows this configuration to be easily exposed via [Zarf Values](https://docs.zarf.dev/ref/examples/values-templating/).  For more information on this topic, see the [Values Guidelines](../guidelines/values-guidelines.md).
```
zachariahmiller, line 46: git diff is kinda messing with me here i think the only other two places you changed are referencing the zarf docs and docs you are adding on values, in which case that seems reasonable assuming we are good with these guidelines living here instead of the uds doc site more generally.

docs/uds-packages/guidelines/values-guidelines.md
```diff
@@ -0,0 +1,238 @@
+# UDS Package Values Guidelines
```
zachariahmiller, line 1: This definitely makes sense until full adopted at which some of this probably belongs in the uds packaging docs in the doc site instead or additionally.

docs/uds-packages/guidelines/values-guidelines.md
Racer159: yeah I think so - except where the negative list is the longer list (which might happen sometimes)

docs/uds-packages/requirements/uds-package-requirements.md
Racer159: IMO all of this needs to make its way to the docs site eventually - Possibly under UDS Core or a new "Packages" section... That's likely a later task though.

docs/uds-packages/requirements/uds-package-requirements.md
Racer159: Not with UDS CLI alone - doesn't preclude bundles but it should be deployable with Zarf first then other tools (not lean into bespoke things other tools do) - will work to clarify.

docs/uds-packages/guidelines/values-guidelines.md
Racer159: they are defaults like a Helm chart values default - if you'd like that we could discuss in Zarf as an issue with more context / examples of what those named defaults would be (right now they compose together into a single file).

Racer159: yes - that is a likely need (Brandt also chimed in offline that he sees that as something Zarf will have) - I can experiment with what that might look like.

docs/uds-packages/guidelines/values-guidelines.md
Racer159: I can include this as a specific note!  Definitely would want a single entrypoint for them.

Racer159: I think I addressed this

docs/uds-packages/requirements/uds-package-requirements.md
Racer159: I think I addressed this

docs/uds-packages/guidelines/values-guidelines.md
Racer159: I added a link to the open issue here - https://github.com/zarf-dev/zarf/issues/4918

docs/uds-packages/guidelines/values-guidelines.md
```diff
@@ -0,0 +1,253 @@
...
+
+When implementing UDS Packages, it is important to consider how [Zarf Values](https://docs.zarf.dev/ref/examples/values-templating/) will be exposed from the package and what the implications are for downstream consumers.  UDS Packages are expected to be deployed in a variety of environments, and the values exposed through to Helm charts should be flexible enough to accommodate different deployment scenarios while also being secure and easy to use.
+
+This means that for any given UDS Package most of the values from the underlying Helm chart should be exposed as Zarf Values with only values that would inherently break the UDS Package (i.e. `image` references, or `ingress` configurations) or compromise the security of the package (i.e. `podSecurityContext`) being restricted.
```
briantwatson:
```suggestion
This means that for any given UDS Package most of the values from the underlying Helm chart should be exposed as Zarf Values. Only values that would inherently break the UDS Package (e.g. `image` references, or `ingress` configurations) or compromise its security (e.g. `podSecurityContext`) should be restricted.
```
nit: on phrasing, breaking in to two sentences might read a little easier. Also I think we want "e.g." here an a few other places in the document as these aren't exhaustive lists but instead examples?  With i.e. I think _"that is"_ or _"in other words"_

docs/uds-packages/guidelines/values-guidelines.md
```diff
@@ -0,0 +1,253 @@
...
+            targetPath: .volumeMounts
+          - sourcePath: .nginx.nodeSelector
+            targetPath: .nodeSelector
+          - sourcePath: .nginx.tolerations
+            targetPath: .tolerations
+          - sourcePath: .nginx.affinity
+            targetPath: .affinity
+```
+
+Within the root `zarf.yaml` the top-level `values` key maps to a `zarf-values.yaml` file that contains the default values for the package and a `zarf-values.schema.json` file that contains the schema for those values.  The schema has specific types for common options (i.e. `replicaCount`), but looser schema for other options (i.e. `affinity`, `tolerations`, `nodeSelector`, etc.) to reduce the complexity of the schema and make it easier to understand and maintain.
```
briantwatson:
```suggestion
Within the root `zarf.yaml` the top-level `values` key maps to a `zarf-values.yaml` file that contains the default values for the package and a `zarf-values.schema.json` file that contains the schema for those values.  The schema has specific types for common options (e.g. `replicaCount`), but looser schema for other options (e.g. `affinity`, `tolerations`, `nodeSelector`, etc.) to reduce the complexity of the schema and make it easier to understand and maintain.
```

briantwatson (commented):
A couple small comments on phrasing.

docs/uds-packages/guidelines/values-guidelines.md
Racer159: yep `i.e.` / `e.g.` is definitely one I get loose on - thanks for the suggestion, went ahead and committed it.

joelmccoy (approved):

docs/uds-packages/guidelines/values-guidelines.md
zachariahmiller: I'm generally concerned about prescribing default values as a whole separate thing that **needs** to be defined. In a lot of cases i dont think we want to be setting default zarf values for exposed values other than what is already the default in the underlying chart (or zarf component helm values) file. Requiring the default values to be filled out seems like adding an additional layer which a bunch of duplication that can potentially lead to drift of the intended config and unintended consequences. 

Furthermore, i think the schema needs to be able to be generated and there is already an issue in zarf for that, which is great, but if it has to be derived from the default values file that is problematic for the same reasons i think having to define all of the defaults if we are exposing is problematic. 

I have a separate concern about how we replace zarf variables that are being used to build helm values. For example, in the GitLab package we build s3 bucket names from a prefix and suffix var:

```
    lfs:
      bucket: ###ZARF_VAR_BUCKET_PREFIX###gitlab-lfs###ZARF_VAR_BUCKET_SUFFIX###
    artifacts:
      bucket: ###ZARF_VAR_BUCKET_PREFIX###gitlab-artifacts###ZARF_VAR_BUCKET_SUFFIX###
    uploads:
      bucket: ###ZARF_VAR_BUCKET_PREFIX###gitlab-uploads###ZARF_VAR_BUCKET_SUFFIX###
```

This doesn't play well with the zarf values feature as i understand it (happy to be wrong) and this is definitely not the only place patterns like this are used. Another that comes to mind is building connection strings.

zachariahmiller: Just noting the issues I've raised here will likely be able to be resolved by a combination of https://github.com/zarf-dev/zarf/pull/4961 and something along the lines of https://github.com/zarf-dev/zarf/issues/4680, which is less of an immediate concern but a concern for post 1.0 zarf.

zachariahmiller (approved):

PR #680, opened 2026-05-21
chore: swap from oras setup task to shell

codyshoffner:
## Description

The [oras-setup](https://github.com/oras-project/setup-oras) action is not often updated and because of this, is often incompatible with the latest Oras CLI versions. Renovate frequently wants to update Oras CLI, which will in-turn break the `oras-setup` task, because it does not often support the latest version. 

This PR is to migrate away from the `oras-setup` task, in favor of creation our own uds-task that will install oras.

Fixes FDRY-112
## Testing
On the dummy package, I bumped releaser and pointed all versions of `uds-common` to this current branch of `ci/oras-setup`, then I published a release. You can see oras successfully setup on the [dummy actions](https://github.com/uds-packages/dummy/actions/runs/26234327144/job/77203110221)

I also created a one-off workflow on dummy to verify oras installs and I verify the [path here](https://github.com/uds-packages/dummy/actions/runs/26234069718/job/77202183376)

## Checklist before merging

### DIFF
--- a/tasks/actions.yaml
+++ b/tasks/actions.yaml
@@ -138,6 +138,21 @@ tasks:
           "https://github.com/defenseunicorns/uds-pk/releases/download/${UDS_PK_VERSION}/uds-pk_${UDS_PK_VERSION}_$(uname -s)_${{ .variables.ARCH }}" \
           && chmod +x /usr/local/bin/uds-pk
 
+  - name: install-oras
+    description: Install ORAS CLI for OCI artifact operations
+    actions:
+      - task: determine-arch
+      - description: Install ORAS CLI
+        env:
+          # renovate: datasource=github-tags depName=oras-project/oras versioning=semver
+          - ORAS_VERSION=v1.3.2
+        cmd: |
+          OS=$(uname -s | tr '[:upper:]' '[:lower:]')
+          curl -fsSL \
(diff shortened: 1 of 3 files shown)


tasks/actions.yaml
```diff
@@ -138,6 +138,21 @@ tasks:
...
           && chmod +x /usr/local/bin/uds-pk
 
+  - name: install-oras
+    description: Install ORAS CLI for OCI artifact operations
+    actions:
+      - task: determine-arch
+      - description: Install ORAS CLI
+        env:
+          # renovate: datasource=github-tags depName=oras-project/oras versioning=semver
+          - ORAS_VERSION=v1.3.2
```
zachariahmiller, line 148: input with default instead of env?

codyshoffner: @zachariahmiller would that be a deviation with how we are installing other deps? 
https://github.com/defenseunicorns/uds-common/blob/ba03e5bec6704a54466a94a3fdee9741b755630b/tasks/actions.yaml#L134

zachariahmiller: i suppose its fine to leave it like this its even implemented differently for GO_VERSION...

zachariahmiller (approved):

PR #701, opened 2026-06-15
ci: change republish job to manual dispatch

codyshoffner:
## Description

- Change republish job to only run on manual dispatch instead of on PRs

## Checklist before merging

### DIFF
--- a/.github/workflows/republish-package.yaml
+++ b/.github/workflows/republish-package.yaml
@@ -4,17 +4,7 @@
 name: Republish-Package
 
 on:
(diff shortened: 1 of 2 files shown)


zachariahmiller (dismissed):

codyshoffner:
The republish job is breaking CI with the new zarf values features that have been added. Bypassing broken CI to turn this job off so we can publish a working version without this overwriting it afterwards.

zachariahmiller (approved):

PR #707, opened 2026-07-01
chore: remove deprecated --no-progress from zarf commands

briantwatson:
## Description

Saw several of the same warning in CI:

`Flag --no-progress has been deprecated, Progress bars and spinners were removed with --log-format=legacy, this flag
  will be removed in a future version of Zarf.`

 I think `--no-progress` can be removed from all zarf commands.

## Checklist before merging


codyshoffner (approved):
Looks good! Thanks Brian!

PR #708, opened 2026-07-03
chore: add a renovate rule to check for versions afer a slash

Racer159:
## Description

This adds a renovate rule to find versions after a slash.  The prefix `# renovate-slash` is used because otherwise it could overlap with the colon rule - this is possible to workaround with ordering but it felt safer to have a namespace.

## Checklist before merging


Racer159:
This is tested here:

https://github.com/Racer159/renovate-playground/pull/2

(note the zarf init 0.78.0 -> 0.80.0 update in README.md and zarf.yaml)

codyshoffner (approved):

PR #720, opened 2026-07-16
feat: add helm linting to lint tasks

### LINKED ISSUE #428: Faster feedback on helm chart issues/formatting standards
### Is your feature request related to a problem? Please describe.

Currently our helm charts are "tested" at deploy time only, and excluded from the minimal `yamllint` checks we do. It would be nice to have a fast failure on the helm charts if they are malformed, and also some standard linting against them.

### Describe the solution you'd like

A few suggestions to run in CI/locally, tied to the current lint tasks likely:
- `helm lint --strict src/**/chart`: this is very basic linting, but would catch small things quickly
- `helm template <chart>`: Good smoke test that the chart will template successfully with default values, but would not validate edge cases with certain values configuration
- `helm template <chart>` + `yamllint` against the templated files: This would be nice to catch weird/wrong formatting, a bit more advanced and longer to run

### Describe alternatives you've considered

I think there are things that might be too far for "smoke testing", such as using https://github.com/helm/chart-testing/tree/main , but could be worth exploring for some of the specific charts we maintain (keycloak/authservice?).

### Additional context

This was spawned out of attempting to use `yamllint` against charts and hitting issues due to the helm template formatting. It would be nice to havelinting/validation run before our longer CI kicks off.

codyshoffner:
## Description

Add Helm linting to our linting tasks, which should help with faster feedback on Helm templating issues.

This adds two new tasks under `lint`:

- **`uds run lint:helm`** runs `helm lint --strict` on charts to catch chart/values issues.
- **`uds run lint:helm-template`** dry-run renders charts to catch template execution errors.

Both discover charts recursively by default, and accept a default `paths` of `.`  so it can be scoped to a specific path.. Wired into the `lint:all` task.

## Testing

Tested locally against several different repositories by pointing the linting tasks at this branch, running the lint commands, and intentionally breaking Helm templates to confirm the failures were caught.

Fixes #428 
## Checklist before merging

**Expected Failure**
failure

**Expected Success**
success


jwgalvin (commented):
I don't see anything that raises any concerns, but will wait to approve for someone with a little more context.

jwgalvin:
Screenshot 2026-07-17 at 11 57 27 AM
Screenshot 2026-07-17 at 12 02 03 PM
Tested with a bad helm chart, and a good one. looks like it does what it says it does.

jwgalvin:
Ok, @codyshoffner got One small nit: if paths resolves to a directory with no Chart.yaml, both tasks exit 0 silently. A typo like --with paths=src/wrong-name would produce a green run with zero charts validated. Might be worth a guard like:

`[ -z "$charts" ] && { echo "No Chart.yaml files found under: $paths" >&2; exit 1; }
```> [ -z "$charts" ] && { echo "No Chart.yaml files found under: $paths" >&2; exit 1; }`

codyshoffner:
@jwgalvin I ended up adding in your suggestion. It will now print out in a yellow text when it cannot find a chart in the specified directory
CleanShot 2026-07-20 at 10 02 57

jwgalvin (approved):
Thanks for adding that.

PR #730, opened 2026-07-23
chore: make the tasks more compatible with maru runner

Racer159:
## Description

This swaps `uds zarf` references to `./zarf` so that `maru` can run more tasks without needing `uds` installed (good for `uds-cli-next` development).

## Checklist before merging


tasks/lint.yaml
```diff
@@ -55,9 +55,12 @@ tasks:
       task_globs:
         description: Globs to match to Maru task files in the repo
         default: tasks.y[a]ml tasks/*.yaml
+      shell_globs:
+        description: Globs to match to shell script files in the repo
+        default: "*.sh **/*.sh **/**/*.sh **/**/**/*.sh **/**/**/**/*.sh **/**/**/**/**/*.sh **/**/**/**/**/**/*.sh **/**/**/**/**/**/**/*.sh"
```
Racer159, line 60: added this so that callers have some control over shell scripts the pull in similar to zarf yamls and tasks - the extra `*` help catch nested findings without needing globstar (which would break in macOS)

codyshoffner:
Should probably update this README https://github.com/defenseunicorns/uds-common/blob/e2d9247dda92c4edf0e2f474ae4d585c24158293/tasks/README.md?plain=1#L19

zachariahmiller:
> Should probably update this README
> 
> https://github.com/defenseunicorns/uds-common/blob/e2d9247dda92c4edf0e2f474ae4d585c24158293/tasks/README.md?plain=1#L19

It still is not required when using uds run directly, however it is worth updating to add that if using `maru` instead of `uds run, a given version of uds and zarf are required

zachariahmiller (approved):

PR #761, opened 2026-09-09
chore: pass version input for commitlint

codyshoffner:
## Description
- Adds an optional node-version input to the reusable commitlint workflow.
- When omitted, existing runner-default behavior is unchanged.

Can be called like the following
```yaml
with:
  node-version: "24"
```

## WHY
- Dependency-operator uses Node 24, but the shared commitlint workflow falls back to Node 22/npm 10. npm fails while installing commitlint in that environment.
This change lets dependency-operator run commitlint with Node 24.
This lets callers run commitlint with their supported Node version.
- Failure Example: https://github.com/defenseunicorns/dependency-operator/actions/runs/34251025675/job/102555029168?pr=128

## TESTING
 https://github.com/defenseunicorns/dependency-operator/pull/131
## Checklist before merging

### DIFF
--- a/.github/workflows/callable-commitlint.yaml
+++ b/.github/workflows/callable-commitlint.yaml
@@ -12,6 +12,11 @@ on:
 
   # This allows other repositories to call this workflow in a reusable way
   workflow_call:
+    inputs:
+      node-version:
+        description: Node.js version to use for commitlint
+        required: false
+        type: string
 
 # Permissions for the GITHUB_TOKEN used by the workflow.
 permissions:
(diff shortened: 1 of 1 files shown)


michalszynkiewicz (approved):

.github/workflows/callable-commitlint.yaml
```diff
@@ -12,6 +12,11 @@ on:
 
   # This allows other repositories to call this workflow in a reusable way
   workflow_call:
+    inputs:
+      node-version:
+        description: Node.js version to use for commitlint
+        required: false
+        type: string
```
jwgalvin, line 19: Is Node 24 specifically required here, or is it working around the npm 10 edgesOut failure caused by installing into the consumer repository? Commitlint 21.2.2 declares Node >=22.12.0. If this workflow is only supported on Node 24, should we make 24 the explicit default and document that instead of allowing arbitrary versions?

jwgalvin (approved):

PR #766, opened 2026-09-15
feat: add support for building and publishing compose bridge packages

willswire:
## Description

Adds Docker Compose as a supported input to the existing package build and publish flow:

- Converts a root `compose.yaml` into a generated Zarf package definition under `build/` before package creation.
- Overlays the release version into `x-uds.metadata.version` during callable publish workflows and publishes using the generated package metadata.
- Pins Docker Compose 5.5.1 in callable test and publish workflows and configures Renovate to track the Compose Bridge transformation image.

This preserves the existing UDS task interfaces, so package repositories based on the UDS Package Template only need to add a `compose.yaml` to use the Compose Bridge build path.

## Checklist before merging

- [x] ADR proposed if making an architectural change to the repo (not applicable)


zachariahmiller:
@willswire Think this makes sense generally, my biggest concern, which i dont have a simple solution for is that this ends up not being tested within this repo. Less of a concern, but still one is wouldnt work with the (also unfortunately untested) gitlab components out of the box since the compose metadata is getting getting overwritten in the workflow.

willswire:
Would you like to see a smoke test added for the building portion, to this PR? I can do that, but I don't think there is publish testing currently for the traditional package approach.

zachariahmiller:
> Would you like to see a smoke test added for the building portion, to this PR? I can do that, but I don't think there is publish testing currently for the traditional package approach.

Yes, but also there is. This repo uses both release please and releaser. It uses release please for itself and releaser to  publish the included test nginx package.

willswire:
Great! I'll add the tests.

willswire:
I added a Compose-based nginx fixture alongside the existing  package and wired it into the existing `test-install` task. Actions on PRs now build both variants for the upstream flavor, so Compose detection, Bridge conversion, and Zarf package creation are exercised before merge.

The existing `publish-release` task also builds and publishes both variants. Actual registry publishing remains release-only, consistent with the traditional package flow.

zachariahmiller:
> I added a Compose-based nginx fixture alongside the existing package and wired it into the existing `test-install` task. Actions on PRs now build both variants for the upstream flavor, so Compose detection, Bridge conversion, and Zarf package creation are exercised before merge.
> 
> The existing `publish-release` task also builds and publishes both variants. Actual registry publishing remains release-only, consistent with the traditional package flow.

okay that seems like a reasonable compromise for now. If this does start getting used regularly I would want to make sure we get coverage on the release even if its something dumb like also publishing nginx-composed.

zachariahmiller (approved):

